PR #74, opened 2026-03-27
Add HTML report generator

pqweraddy:
### Summary

This PR adds an HTML report generator that can convert REPORT.md into styled HTML file and modifies session instructions to a create a HTML report automatically.

### What's included: 

1. New file: src/utils/generate_html.py — Converts a workspace's REPORT.md and figures into a self-contained, styled HTML file. Key features:

* Syntax-highlighted fenced code blocks via Pygments (github-dark theme)
* MathJax support for LaTeX math in markdown
* Experiment metadata (domain, author) pulled from .neurico/idea.yaml if present

ScreenRecording2026-03-27at11 58 23-ezgif com-video-to-gif-converter (2)

2. New file: generate_html.py — Thin entry-point wrapper at the repo root so the script can be invoked as python generate_html.py <workspace> without referencing the full module path.

3. Modified: templates/agents/session_instructions.txt — Added a completion step instructing the research agent to run generate_html.py after writing REPORT.md, so every session automatically produces a HTML report along with REPORT.md.
    
### Test plan
* Run python generate_html.py <workspace> on a workspace with a REPORT.md and figures; verify REPORT.html is produced and renders correctly in a browser
* Confirm embedded images display without external file references
* Confirm the session agent produces report.html at the end of a research run

### DIFF
--- a/generate_html.py
+++ b/generate_html.py
@@ -1,0 +1,4 @@
+"""Thin wrapper so you can run `python generate_html.py <workspace>` from the repo root."""
+from src.utils.generate_html import main
+if __name__ == "__main__":
+    main()
--- a/src/utils/generate_html.py
+++ b/src/utils/generate_html.py
@@ -1,0 +1,635 @@
+"""
+HTML Generator
+Converts REPORT.md + figures into a self-contained styled HTML blog post.
+
+Usage:
+    python generate_html.py <workspace_dir>
+    python generate_html.py <path/to/REPORT.md>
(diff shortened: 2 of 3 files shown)


laoliu5280 (changes requested):
Can you please reflect to the latest template changes, especially with multiple domain specific template overrides? If we want to support html report by default, we will need to handle different domain template overrides gracefully too.

Also why is generate_html.py under project root? Can you please make it better organized? If this is meant to be a quick script, put it somewhere in scripts/ (you can create a scripts folder).

PR #124, opened 2026-06-19
Add modal-training and modal-vllm compute-backend skills

AndrewRqy:
## Summary

Adds two NeuriCo skills that let experiments offload GPU work to [Modal](https://modal.com):

- **`modal-training`** — data prep, LoRA SFT, full SFT, eval
- **`modal-vllm`** — vLLM HTTPS serving

Both implement a strict **leave-no-trace lifecycle**: per-experiment Modal environments are created at register, all reproducibility artifacts are pulled back to the workspace, and the environment is destroyed at teardown (cascading volumes, apps, and secrets). Nothing persists Modal-side after a clean run, by design — this is what keeps experiment throughput scalable.

Agents discover compute backends by a **frontmatter tag** (`compute-backend`) rather than by name, so future backends (Slurm, Lambda, local-GPU, …) drop in without prompt edits.

## Why now

Many ideas require GPU training beyond what the local Docker container provides. Without a clean offload story, agents either fail silently or accumulate cloud cruft that bottlenecks future runs. This PR is the foundation for moving NeuriCo's compute ceiling up while preserving the "workspace owns everything" reproducibility contract.

### `modal-training` skill

| File | Purpose |
|---|---|
| `SKILL.md` | Agent-facing entry point, prerequisites, decision tree |
| `references/{modal_basics,training_recipes,lifecycle,cost_guide}.md` | Reference docs, mandatory reading for customization |
| `scripts/lifecycle.py` | `register` / `pull_all` / `teardown` — the load-bearing primitives |
| `scripts/new_modal_app.py` | Scaffolder; produces a runnable `modal_*.py` from a template |
| `scripts/modal_sweep.py` | Invoked by the pipeline orchestrator at workspace teardown |
| `scripts/check_modal_setup.py` + `_doctor_checks.py` | Setup doctor (CLI/auth/secrets/disk/probe) with `--json` for agents |
| `scripts/templates/*.tmpl` | Parameterized Modal apps (LoRA train, data prep, eval) |

### `modal-vllm` skill

Same shape, slimmer surface. `lifecycle.py` is a thin wrapper over `modal-training`'s lifecycle that adds the deploy / `app stop` / endpoint-capture / redact-to-artifacts steps.

### Pipeline / runtime wiring

- `pyproject.toml`: adds `modal>=1.4` to core deps so the `modal` CLI ships in the venv.
- `docker/run.sh`: mounts `~/.modal.toml` read-only into the container (same pattern as Claude / Codex / Gemini credentials), plus an entry in the credential-check printer.
- `src/core/pipeline_orchestrator.py`: adds `_modal_sweep_if_used()` in the existing `finally` block. Sentinel-gated — non-Modal runs pay ~50µs (a single `Path.exists()`).
- `templates/agents/session_instructions.txt`: new "COMPUTE BACKENDS BEYOND THE LOCAL CONTAINER" section that tells the agent to discover by tag, never by provider name.
- `templates/agents/autoresearch_proposer.txt`: shorter "COMPUTE BUDGET" section so the proposer scopes around compute availability.
- `templates/README.md`: documents the `compute-backend` tag convention for future backends.

### Manifest-driven artifact pulls (no hardcoded training conventions)

`lifecycle.pull_all()` has zero opinions about what a run produces. Each template declares its own pull manifest in the `register()` call:

\`\`\`python
PULL_MANIFEST = [
    {\"from_volume\": CKPT_VOLUME, \"from\": \"/run_config.json\",
     \"to\": \"artifacts/run_config.json\", \"required\": True},
    {\"from_volume\": CKPT_VOLUME, \"from\": \"/final\",
     \"to\": \"artifacts/final\", \"is_dir\": True, \"required\": True},
]
lifecycle.register(EXP_ID, volumes=[...], pull_manifest=PULL_MANIFEST, ...)
\`\`\`

A new training shape (DPO, full SFT, quantize+upload) needs zero lifecycle edits — just a manifest. Multi-stage runs (prep + train) merge their manifests into the sentinel automatically.

### Per-env secrets minted from local env vars

Modal secrets are environment-scoped — a secret in `main` is invisible to `neurico-<EXP_ID>`. `register(required_secrets={...})` mints them per-env from local env vars and they cascade-delete with the env at teardown. The scaffolder accepts repeatable `--secret NAME=ENV_VAR[,ENV_VAR2]` for multi-credential setups:

\`\`\`bash
python new_modal_app.py lora-train --exp-id foo \\
    --secret wandb-secret=WANDB_API_KEY \\
    --secret openai-secret=OPENAI_API_KEY \\
    --secret s3-creds=AWS_ACCESS_KEY_ID,AWS_SECRET_ACCESS_KEY \\
    --out src/modal_train.py
\`\`\`

### Defense-in-depth teardown

Three layers, each gated on the previous:

1. Template's `main()` `finally` block calls `pull_all()` then `teardown()`
2. Skill's `teardown()` refuses unless `pull_complete=True` (preserves data on failure)
3. Pipeline orchestrator's `_modal_sweep_if_used()` runs at workspace teardown — catches scripts that crashed before their own `finally` block

The user can always recover with `lifecycle.py pull --exp-id <id>` and re-attempt the chain.

### Tag-based agent discovery

Agent prompts never name a provider. They list `.claude/skills/*/SKILL.md`, find frontmatter tagged `compute-backend` + `training` / `serving`, and pick. New backends drop in without prompt changes.

## What users have to set up

The full footprint to use this:

| Setup tier | What to do | When |
|---|---|---|
| 0 — install neurico | `uv sync` or pull the Docker image | Once |
| 1 — Modal auth | `modal token new` on host | Once per machine |
| 2 — gated HF models | Add `HF_TOKEN=hf_xxx` to `neurico/.env` | Once, only if needed |
| 3 — other secrets (W&B, OpenAI judge, S3, ...) | Add env vars to `neurico/.env` | Once per credential type, only if needed |
| 4 — vLLM serving | Modal dashboard → mint proxy-auth token pair | Once per deployed endpoint, only if using vllm |

That's it. The doctor (`check_modal_setup.py --probe`) surfaces any missing piece with the exact fix command.

## Live test results

End-to-end exercised on real Modal:

| Scenario | Result | Spend |
|---|---|---|
| Data prep CPU run end-to-end | ✓ manifest pull + clean teardown | ~\$0.05 |
| LoRA training on L4 (Qwen2.5-0.5B, 5 steps) | ✓ 13 artifacts pulled, env destroyed | ~\$0.20 |
| vLLM deploy → capture-endpoint → redact → app stop → env delete | ✓ live JSON cleared, redacted copy preserved | ~\$0.30 |
| Training crash mid-finalize → orchestrator sweep preserves env | ✓ defense-in-depth held; user recovery via \`lifecycle.py pull\` precisely surfaced the missing artifact | ~\$0.20 |

Bugs surfaced during live testing and fixed in this PR:

- Module name collision (`import lifecycle as base` returning the cached vllm module)
- Modal CLI required `-y` for non-interactive `app stop`
- `modal volume get` "not found" error phrasing variants
- Directory pull semantics (destination must end with `/`, leaf must match remote leaf)
- HF Trainer `save_steps` interaction with short runs (template now copies `trainer_state.json` to a stable path)

## Backward compatibility

- **Non-Modal experiments**: zero behavior change. The orchestrator hook is gated on a sentinel file; a single `Path.exists()` per pipeline finally block.
- **No existing skills modified**: all changes are additive at the file level.
- **No new required dependencies for non-Modal users**: `modal` adds ~7 MB to the venv and is otherwise inert until imported.

## Test plan

- [x] All Python compiles with `-W error::SyntaxWarning`
- [x] `resolve_secrets()` unit tests (4 valid + 6 malformed inputs)
- [x] All 4 templates render + AST-parse in `--no-hf-secret`, default, and multi-secret modes
- [x] Sweep hook smoke tests (mocked subprocess, 4 assertions) pass
- [x] Live data-prep run end-to-end
- [x] Live LoRA training run end-to-end
- [x] Live vLLM deploy → capture → redact → teardown
- [x] Live failure-recovery: crash mid-finalize → orchestrator preserves env → user recovers via `lifecycle.py pull`
- [x] Doctor handles both auth sources (`~/.modal.toml` mount, env vars) and surfaces the right fix for each
- [x] No leftover Modal envs/apps/secrets/volumes after the full test sweep

## Out of scope (follow-ups)

- Intermediate checkpoint metadata pull (currently the manifest captures only the final checkpoint + trainer_state)
- Modal run / function-call IDs in `modal_run.json` (currently null; needs `modal.current_function_call_id()` integration)
- Doctor exit-code + lifecycle CLI unit tests (live coverage stands in for now)


Frankbest18 (changes requested):
Thanks for the PR. I reviewed the Modal lifecycle implementation and noticed a few issues worth further attention.

1. LoRA potentially has no data

`modal_data_prep.py.tmpl` writes `/data/train.jsonl` and `/data/val.jsonl` into the Modal data volume, pulls local copies back into the workspace, and then calls `lifecycle.teardown()`.

The LoRA template later expects those files to exist in the Modal volume at `/data/train.jsonl` and `/data/val.jsonl`, but I do not see an upload/restore path that pushes the workspace copies back into Modal before training.

2. vLLM teardown can be blocked after following the generated instructions

The vLLM scaffolder prints a flow of `modal deploy` -> `capture-endpoint` -> `teardown`

But `capture_endpoint()` only writes `.neurico/modal_endpoint.json` and sets `endpoint_captured=True`. It does not call `pull_all()` or mark `pull_complete=True`.

Then vLLM `teardown()` delegates to the base lifecycle, which refuses teardown when `pull_complete=False`.

Since vLLM uses `modal deploy`, this can leave a persistent serving app running/billing if the user follows the generated instructions exactly.

3. Orchestrator sweep assumes Claude-style skill layout

`_modal_sweep_if_used()` looks only for `.claude/skills/modal-training/scripts/modal_sweep.py`

That means Modal cleanup depends on the skills being installed under `.claude/skills`, even though the pipeline supports other providers such as Codex and Gemini.

AndrewRqy:
Thanks for the careful read:) all three confirmed and fixed in three new commits on top of e4e37a7. One commit per review point.
  
  1. LoRA potentially has no data — 2e5944a
  Confirmed: data_prep tears its env down, which cascade-deletes the data volume; a chained LoRA run under the same EXP_ID was
  hitting a fresh empty volume.
  - Added upload_to_volume(exp_id, volume, src_workspace_rel, dest_volume_path) to modal-training/scripts/lifecycle.py.
  Sentinel-aware (validates the volume was registered to this env), checks the local source exists, raises with a precise error if
  the chain is broken.
  - modal_lora_train.py.tmpl's main() now re-uploads data/{train,val}.jsonl from the workspace before train.remote(). The workspace
  data/ directory is the source of truth between stages.
  - Documented the chain-ownership rule in references/training_recipes.md so any future chained stage reaches for the same helper.

  2. vLLM teardown can be blocked after following the generated instructions — ab4a39a
  Confirmed: capture_endpoint() only flips endpoint_captured=True, and base.teardown() refuses without pull_complete=True, so a user
  following the printed flow exactly got the deployed app stuck running and billing.
  - new_modal_app.py now prints a 4-step flow: modal deploy → capture-endpoint → lifecycle.py pull → lifecycle.py teardown.
  - lifecycle.teardown() self-heals as defense-in-depth: when endpoint_captured=True && pull_complete=False && force=False, it runs
  pull_all() first. A failed self-heal raises and leaves the env alive so the user can recover via lifecycle.py pull. force=True
  (used by the orchestrator sweep when no pull is expected) bypasses self-heal.
  - SKILL.md step 4 and references/lifecycle.md updated to show the 4-step flow and document the self-heal escape hatch.

  3. Orchestrator sweep assumes Claude-style skill layout — 6efbd01
  Confirmed and agreed — even though runner.py currently mirrors skills to all three (.claude/.codex/.gemini), cleanup correctness
  shouldn't depend on Claude's directory being populated.
  - _modal_sweep_if_used now takes provider and walks [preferred provider's skill dir, …other providers' dirs], taking whichever has
  modal_sweep.py. If none does, it warns naming all three searched paths and lets the user clean up manually.
  - Behavior on Claude is unchanged. Codex/Gemini-only setups now sweep correctly.

Let me know if you'd rather see them as committed tests/ files than the inline smokes I ran locally, happy to add either way.

Frankbest18 (changes requested):
Thanks for the follow-up fixes.

1. The vLLM cleanup flow looks addressed now.
2. Since .claude, .codex, and .gemini are all generated in the workspace at the start, relying on one of them should not be a blocker.

I only see one remaining issue:

LoRA re-upload appears to target the wrong volume path

The data-prep stage writes files inside the mounted volume at /data/train.jsonl and /data/val.jsonl, and the pull manifest reads them from the volume root as /train.jsonl and /val.jsonl.

The new LoRA re-upload code uploads the local workspace copies to /data/train.jsonl and /data/val.jsonl inside the volume. Since the volume itself is mounted at /data in the remote container, this likely creates /data/data/train.jsonl and /data/data/val.jsonl when mounted.

But the training function still reads /data/train.jsonl and /data/val.jsonl, so it may still miss the restored data.

This should be a small fix if confirmed: the re-upload destination should likely be /train.jsonl and /val.jsonl inside the volume.

Other than this minor point, the follow-up commits look great to me.

AndrewRqy:
Thanks a lot for the follow ups! Changed the re-upload destinations in modal_lora_train.py.tmpl to /train.jsonl and /val.jsonl, and called out the volume-root-vs-container-path distinction in the upload_to_volume() docstring and in references/training_recipes.md so the same trap doesn't happen in the next chained stage.

Frankbest18 (approved):

PR #127, opened 2026-06-21
Added `--provider agy` (Google Antigravity CLI) support to neurico ac…

Chenxi1818:
Added `--provider agy` (Google Antigravity CLI) support to neurico across the Python provider configs and Docker layer, with agy reusing Gemini's credentials. The dry-run confirmed command building, argument parsing, and a real agy launch all work.

**Reason for the action:**
47b9c5f61ee8426f300c6ab28b7c5897

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -69,6 +69,11 @@ RUN uv python install 3.12
 RUN curl -fsSL https://claude.ai/install.sh | bash \
     && cp /root/.local/bin/claude /usr/local/bin/claude
 
+# Install Google Antigravity (agy) via native installer (installs to ~/.local/bin).
+# agy reuses Gemini's OAuth credentials (~/.gemini/oauth_creds.json), so no extra cred dir.
+RUN curl -fsSL https://antigravity.google/cli/install.sh | bash \
+    && cp /root/.local/bin/agy /usr/local/bin/agy
+
 # Install Codex and Gemini via npm (no native Linux installer available)
 RUN npm install -g @openai/codex \
     && npm install -g @google/gemini-cli
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -90,6 +90,7 @@ validate_env() {
     # OPENAI_API_KEY: Required for IdeaHub integration and paper-finder
     # GITHUB_TOKEN: Required for GitHub repo creation
     # Note: Claude/Codex/Gemini CLIs use OAuth credentials from ~/.claude, ~/.codex, ~/.gemini
+    # (Antigravity/agy reuses ~/.gemini/oauth_creds.json)
 
     if [ -n "$OPENAI_API_KEY" ]; then
         echo -e "  ${GREEN}[OK]${NC} OPENAI_API_KEY configured (IdeaHub, paper-finder)"
@@ -222,7 +223,7 @@ show_help() {
     echo -e "    ${GREEN}python /app/src/core/runner.py <idea_id> [options]${NC}"
     echo ""
     echo "  Options for runner.py:"
-    echo "    --provider {claude|codex|gemini}  AI provider (default: claude)"
+    echo "    --provider {claude|codex|gemini|agy}  AI provider (default: claude)"
     echo "    --full-permissions                Skip permission prompts"
     echo "    --no-github                       Run locally without GitHub"
     echo "    --timeout SECONDS                 Execution timeout (default: 3600)"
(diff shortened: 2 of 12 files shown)


AndrewRqy (changes requested):
1. templates/manager/tools.yaml enum missing agy

  - What: Line 19, provider: enum is [claude, codex, gemini].
  The interactive manager's run_agent tool schema rejects agy at
  the LLM-tool-call layer even though argparse choices accepts
  it everywhere else.
  - Fix: Add agy to the enum.

  2. agy -p is hitting the default 5-minute --print-timeout 
  mid-task

  - What: PR ships agy -p --dangerously-skip-permissions, which
  uses agy's default --print-timeout=5m0s. A live
  resource_finder run ran for exactly 308.9s before agy aborted
  with "Error: timed out waiting for response" — it had
  downloaded 4 papers and written pyproject.toml/uv.lock but
  hadn't synthesized literature_review.md/resources.md yet.
  Other stages (experiment_runner at 45-90 min, paper_writer at
  60 min) would blow through 5 min instantly.
  - Fix: Append --print-timeout to the agy command across all
  six dispatch sites (runner.py, pipeline_orchestrator.py,
  agent_runner.py, paper_writer.py, comment_handler.py,
  resource_finder.py). Scale it to match each stage's existing
  timeout argument — e.g. --print-timeout 45m for
  resource_finder, --print-timeout 90m for experiment_runner.
  Easy passthrough.

  3. Documented OAuth filename is wrong; the wizard's "already 
  configured" check is dead code

  - What: Actual path agy writes is
  ~/.gemini/antigravity-cli/antigravity-oauth-token, not
  ~/.gemini/oauth_creds.json as the PR comments and the
  agy_status check at docker/run.sh:1337 claim. So [ -s 
  "$HOME/.gemini/oauth_creds.json" ] never matches — the wizard
  always shows agy as un-configured even after a successful
  login.
  - Fix: Update the status check, the Dockerfile comment
  (:73-74), and run.sh:265's comment to the real path. The bind
  mount itself is fine since the whole ~/.gemini/ subtree is
  mounted.

  4. cmd_login short-circuits on Claude credentials regardless 
  of provider

  - What: Pre-existing bug at docker/run.sh:777. ./neurico login
  agy exits early if ~/.claude/.credentials.json exists,
  without ever launching the agy login flow. The PR added the
  agy instruction line just above it, so the agy onboarding path
  is now actively impacted.
  - Fix: Make the short-circuit provider-aware — match the
  credential file to $1, not Claude unconditionally.

  5. setup_login_provider false positive when Gemini is already 
  logged in 

  - What: Lines 1108-1119. The "already configured" check is [ 
  -d "$host_dir" ] && [ "$(ls -A "$host_dir")" ]. With $host_dir
  = ~/.gemini, any Gemini auth (or stray file) triggers
  "Antigravity credentials already configured — re-login? 
  [y/N]", default N. Users press Enter and quietly skip the
  actual agy login.
  - Fix: Add a 5th parameter existence_test_path and check that
  file specifically; for agy pass
  ~/.gemini/antigravity-cli/antigravity-oauth-token.

  6. Pipeline reports success when agy produced no outputs

  - What: Because the agy run currently exits 0 even when none
  of the expected outputs exist, the log shows "✅ Agent 
  execution completed successfully!" immediately followed by "⚠️ 
  Resource finder stage failed!". These two messages refer to
  different things (CLI exit vs. output validation) and read as
  contradictory.
  - Fix: When the provider has no transcript-flag support
  (currently just agy), gate the success message on output
  validation rather than exit code. Likely resolves itself once
  item #2 is in, but the guard is still worth having.

  7. agy has no streaming-JSON transcript flag

  - What: TRANSCRIPT_FLAGS lacks an agy entry (PR notes this
  with a comment). So agent_log.jsonl is empty for agy runs —
  operators have no per-turn detail to debug failures with.
  - Fix: Either (a) add a clear note in the README that
  --provider agy produces no transcript, or (b) follow up with a
  wrapper that parses agy's stdout into JSONL. (a) is fine for
  v1.

  8. Stale provider triples in docs/docstrings
  
  - What: Eight Python docstrings ((claude, codex, gemini)),
  README, INTERACTIVE_MODE_GUIDE, ARCHITECTURE,
  DOCKER_PERMISSIONS, WORKFLOW docs, plus
  clawskill/clawhub.json:47's supported_providers list still
  omit agy.
  - Fix: grep -rn "claude.*codex.*gemini" sweep. Follow-up
  commit.

PR #131, opened 2026-06-22
feat(interactive): concurrency-safe merge-on-save for the shared world model

jingxuangu:
## Summary

**PR 4 of the v3 world-model series.** The world model is becoming **shared memory** that the manager and (soon) research sub-agents all write. The old read-modify-write `_save` was **last-writer-wins** and would clobber a concurrent writer's record. This PR makes saves **merge** (CRDT-lite) under an advisory lock.

> Stacked on [PR #130](https://github.com/ChicagoHAI/NeuriCo/pull/130) → #129 → #128. This is the gated/forward-looking PR from the plan — single-writer behaviour is unchanged today; it makes the store correct the moment a second writer appears.

## How it works

- **`_save` → read-merge-write under a best-effort `flock`.**
  - id-keyed lists (`hypotheses`/`experiments`/`findings`/`decisions`) **union by id**, newest-by-timestamp wins on a tie.
  - `incidents` union by `(kind, detail)`; `sections` merge by key (newest `updated_at`).
  - Manager-owned scalars (`narrative`/`current_best`/`crux`/`open_questions`/`panel_layout`) are re-asserted **only when this instance changed them** (dirty tracking) — so a field we didn't touch is never overwritten with our stale value.
- **`_next_id`** folds in the on-disk ids so a freshly minted id doesn't collide with another writer's record we haven't merged yet.
- **Migration save is a plain overwrite** (`merge=False`): we just loaded that file, so merging against it would re-add the pre-migration records (e.g. an id-less flat finding) as duplicates.
- **No behaviour change for a single writer** (merge against our own last write is a no-op); `flock` no-ops where `fcntl` is absent (e.g. Windows).

## Documented residual (the next decision)

Id minting is *best-effort* unique (mints against the on-disk ids). Two writers that mint the same id from stale views **before either saves** can still collide. The real fix is a **globally-unique id scheme** — author-scoped (`F-runner-3`) or uuid — which is entangled with how the sub-agent writers (owned by others) mint ids. That's the next step, called out so we decide it together. Until then the realistic case (writers touching different records, or minting against the latest file) is collision-free.

## Testing

- **+5 concurrency tests** (23 total, all passing):
  - merge-not-clobber: a stale writer's save folds in another's record.
  - scalar non-clobber: an untouched `crux` survives a stale writer's save.
  - LWW dirty scalar: an explicitly-set `narrative` wins.
  - cross-instance id minting: F1 then F2, no collision.
  - **8-thread incident race** (barrier-synced) — every incident survives the `flock` + merge. Green across 10 repeated runs.
- `py_compile` clean across the four interactive modules.

## Scope

`src/interactive/research_state.py` + tests — within PR #113's footprint.

## Series complete (core)

#128 data model · #129 manager authoring · #130 whiteboard · **#131 concurrency**. Remaining future work (separate, owned elsewhere): the human-interactor subagent (`should_engage`), the review subagent (`importance`), the sub-agent writers (+ the unique-id scheme above).

### DIFF
--- a/src/interactive/research_state.py
+++ b/src/interactive/research_state.py
@@ -37,9 +37,15 @@ from __future__ import annotations
 
 import json
 import os
+from contextlib import contextmanager
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional
+
+try:  # POSIX advisory file locks; absent on some platforms (e.g. Windows).
+    import fcntl
+except ImportError:  # pragma: no cover - platform dependent
+    fcntl = None
 
 SCHEMA_VERSION = 3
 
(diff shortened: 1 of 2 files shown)


AndrewRqy (commented):
Problem worth fixing in PR #131:

  When N writers race against a freshly-empty store, id-keyed lists (findings, decisions, hypotheses) lose up to N-1 records per write — stress test
  shows 1/8 survivors every trial under 8-thread contention.

  Why it happens: _next_id mints ids by reading what's currently on disk. When all N threads acquire the lock-read-merge-write sequence starting from
  an empty store, each one independently sees zero existing findings and mints F1. The merge unions by id, so N records with the same id F1 collapse
  into one — last writer's F1 overwrites all prior F1s. Incidents survive because they union by (kind, detail) content, not by id, so there's no id
  collision to overwrite on. The fix is globally-unique ids (UUID or a monotonic counter that doesn't reset to F1 on an empty view).

jingxuangu:
> Problem worth fixing in PR #131:
> 
> When N writers race against a freshly-empty store, id-keyed lists (findings, decisions, hypotheses) lose up to N-1 records per write — stress test shows 1/8 survivors every trial under 8-thread contention.
> 
> Why it happens: _next_id mints ids by reading what's currently on disk. When all N threads acquire the lock-read-merge-write sequence starting from an empty store, each one independently sees zero existing findings and mints F1. The merge unions by id, so N records with the same id F1 collapse into one — last writer's F1 overwrites all prior F1s. Incidents survive because they union by (kind, detail) content, not by id, so there's no id collision to overwrite on. The fix is globally-unique ids (UUID or a monotonic counter that doesn't reset to F1 on an empty view).

You're right that this was the one case contradicting the PR's title: incidents survived the 8-thread race only because they union by (kind, detail), while the id-keyed lists collapsed N×F1 into one survivor.

Rather than a global UUID or a shared monotonic counter (both of which would entangle us with how the sub-agent writers mint ids), I namespaced ids per writer: each ResearchState instance gets a random writer token, so ids are now F-<writer>-<n> (e.g. F-a0916b48-1). Two writers racing from an identical empty view mint different ids → the union keeps both. It's your "counter that doesn't reset to F1," made collision-proof by the per-instance namespace, with zero cross-writer coordination — each writer only needs uniqueness, not a meaningful author name. When the sub-agent-writer teams land, they get this for free.

Added the missing test — your exact repro: 8 threads racing to add a finding from a freshly-empty store, all 8 survive with unique ids (green across 10 reruns). It fails on the old scheme.

One trade-off worth naming: ids lose human-readable sequencing (F1 → F-a0916b48-1). That's intentional and cheap — ordering is carried by ts/sequence, and the id is purely a foreign key for decision.finding — but flagging it in case the whiteboard/demo wants prettier ids, in which case the alternative is minting sequential ids inside the save lock (more code, no visible change).

PR #132, opened 2026-06-23
feat: add runtime compute backend selection and DSI Slurm skill

Frankbest18:
## Summary

This PR builds on [PR #124](https://github.com/ChicagoHAI/NeuriCo/pull/124). It keeps PR #124’s Modal skills and compute-backend guidance, while adding runtime backend selection and a new DSI Slurm compute path.

The new backend is implemented as a dsi-slurm skill and presented to agents as dsi-cluster. It provides guidance for running NeuriCo experiment workloads on the University of Chicago Data Science Institute Slurm cluster.

Compute backend selection is controlled at runtime through a CLI flag:

```bash
--compute-backend local
--compute-backend dsi-slurm
--compute-backend modal
```

When local is selected, NeuriCo does not inject Slurm or Modal-specific guidance and does not create any remote workspace.

The intended model for using `dsi-slurm` for computation is:

```text
runtime selects compute backend
→ NeuriCo attaches backend to the in-memory run state
→ relevant agents receive backend guidance only when needed
→ local workspace remains the source of truth for scoring, reporting, comments, and paper writing
```

## CLI behavior

New runtime flag:

```bash
--compute-backend {local,dsi-slurm,modal}
```

Examples:

```bash
./neurico run <id>
./neurico run <id> --compute-backend local
./neurico run <id> --compute-backend dsi-slurm
./neurico run <id> --compute-backend modal
```

Comment mode and standalone agent runner support the same backend selection.

If the flag is omitted, NeuriCo uses `local`.

### `local`

Default behavior.

No Slurm prompt is injected. No Modal prompt is injected. No remote workspace is created. Existing local execution remains unchanged.

### `dsi-slurm`

NeuriCo conditionally injects DSI Slurm guidance into experiment/comment/AutoResearch prompts.

Agents are instructed to use the provider-specific skill path:

```text
.<provider>/skills/dsi-slurm/SKILL.md
```

Examples:

```text
.codex/skills/dsi-slurm/SKILL.md
.claude/skills/dsi-slurm/SKILL.md
.gemini/skills/dsi-slurm/SKILL.md
```

The agent must use the runtime-provided remote workspace, submit work through Slurm, and copy required outputs back to the local workspace before reporting success.

All cluster access is through the configured SSH alias:

```bash
login.ds
```

NeuriCo does not require the agent to construct a username-qualified hostname such as `<user>@...`.

### `modal`

Modal/backend guidance is emitted only when `--compute-backend modal` is selected.

## Prerequisites for `dsi-slurm`

Using `--compute-backend dsi-slurm` assumes the human user has already prepared the access that NeuriCo agents cannot create themselves.

Required human prerequisites:

- the user has an active account on `dsi-cluster`
- the machine running NeuriCo has a configured SSH alias `login.ds`
- `ssh login.ds` works non-interactively enough for agent-run commands
- `rsync` over `login.ds` works from the machine running NeuriCo
- the user has permission to submit Slurm jobs on the cluster

Everything else is job-specific setup. The agent may inspect available partitions, storage, modules, Conda/Mamba, Python packages, and GPUs, and may create project-local environments or scripts when appropriate. If a job requires access or permissions the user has not prepared, the agent would report that missing prerequisite instead of guessing or falling back to another backend.

## dsi-cluster runtime workspace lifecycle

When `--compute-backend dsi-slurm` is selected, NeuriCo runtime creates a remote workspace through `login.ds` before the experiment/comment stage starts.

The runtime writes:

```text
.neurico/dsi_slurm_remote_workspace.json
```

with a contract like:

```json
{
  "backend": "dsi-slurm",
  "ssh_host": "login.ds",
  "workspace_name": "<local workspace name>",
  "remote_root": "$HOME/neurico_workspaces/<local workspace name>",
  "rsync_remote_root": "login.ds:~/neurico_workspaces/<local workspace name>/"
}
```

The runtime owns:

```text
create remote workspace
→ expose remote workspace metadata
→ remove remote workspace after the stage
```

The agent owns:

```text
sync needed files
→ submit Slurm jobs
→ monitor safely
→ copy required outputs back to local workspace
→ copy job evidence back to local dsi-slurm-artifacts/
```

The remote workspace is intentionally not reused if it already exists, to avoid stale artifacts contaminating a run.

## dsi-slurm skill

This PR adds/refines:

```text
templates/skills/dsi-slurm/SKILL.md
```

The skill treats `dsi-cluster` as compute, not artifact storage.

Core contract:

```text
local NeuriCo workspace = source of truth
dsi-cluster = compute backend
```

The skill covers:

- using the configured `login.ds` SSH alias
- avoiding heavy work on login nodes
- using Slurm for compute-node work
- using the runtime-provided remote workspace
- safe monitoring without rapid repeated SSH polling
- copying outputs back to the local workspace
- recording per-job Slurm evidence
- handling cancellation, failure, and recovery
- avoiding fallback to another backend unless the user reruns with a different backend

Each Slurm job records local evidence under:

```text
dsi-slurm-artifacts/
  <JOB_ID>/
    slurm_<JOB_ID>.out
    slurm_<JOB_ID>.err
    job.json
```

`job.json` uses this compact schema:

```json
{
  "job_id": "Slurm job or array ID returned by sbatch --parsable",
  "status": "completed, failed, cancelled, timeout, preempted, or unknown",
  "slurm_state": "raw Slurm accounting state such as COMPLETED",
  "exit_code": "raw Slurm exit code such as 0:0",
  "script": "sbatch script path relative to the workspace root",
  "description": "short purpose of this job",
  "submitted_at_utc": "UTC timestamp recorded after submission",
  "recorded_at_utc": "UTC timestamp recorded after final state lookup"
}
```

## Prompt injection behavior

Backend guidance is conditional.

### Experiment runner

Receives backend guidance only when a non-local backend is selected.

For `dsi-slurm`, the experiment agent is told to read the provider-specific skill path and use the runtime-provided remote workspace.

### Comment handler

Receives `dsi-slurm` guidance when `--compute-backend dsi-slurm` is selected.

The runtime also creates and removes the remote workspace around comment-mode execution.

### AutoResearch proposer

Receives backend guidance when `--compute-backend dsi-slurm` is selected, so proposals can preserve the correct cluster execution contract for the comment agent.

### Added files

- `src/core/compute_backend.py`
  - defines valid runtime backends
  - normalizes missing backend to `local`
  - attaches backend to in-memory run state
  - exposes backend lookup helpers

- `src/core/dsi_slurm_remote.py`
  - creates dsi-cluster remote workspaces through `login.ds`
  - writes `.neurico/dsi_slurm_remote_workspace.json`
  - removes the remote workspace after the stage
  - no-ops unless backend is exactly `dsi-slurm`

### Modified files

- `src/core/runner.py`
  - adds `--compute-backend`
  - prints selected backend
  - attaches runtime backend before run/comment execution

- `src/core/agent_runner.py`
  - adds `--compute-backend`
  - passes backend into standalone agent execution
  - creates/removes dsi-cluster remote workspace for `dsi-slurm`
  - passes remote workspace environment variables to the agent

- `src/core/pipeline_orchestrator.py`
  - creates/removes dsi-cluster remote workspace around experiment execution
  - passes remote workspace environment variables to agent subprocesses
  - archives DSI Slurm artifacts only for `dsi-slurm`

- `src/agents/comment_handler.py`
  - wraps comment handler execution with dsi-cluster remote workspace lifecycle when needed
  - passes remote workspace metadata via environment variables

- `src/agents/autoresearch_proposer.py`
  - injects proposer guidance for `dsi-slurm`

- `src/templates/prompt_generator.py`
  - emits backend prompt sections only when needed
  - omits backend guidance for local runs
  - injects DSI Slurm guidance for `dsi-slurm`
  - keeps Modal/backend guidance conditional

- `templates/skills/dsi-slurm/SKILL.md`
  - defines the DSI Slurm workflow and runtime/agent responsibility split

## Design notes

- Backend selection is runtime state.
- Missing backend means `local`.
- Local runs do not see Slurm or Modal instructions.
- The DSI Slurm skill is provider-aware through paths like `.codex/skills/dsi-slurm/SKILL.md`.
- The skill avoids hard-coding one provider’s skill root.
- Cluster access uses the configured `login.ds` SSH alias.
- `dsi-cluster` is compute only, not artifact storage.
- Local workspace files remain authoritative for scoring, reports, comments, and paper writing.
- Runtime creates/removes the remote workspace.
- Agent handles job execution and copy-back because it knows which outputs the experiment generated.
- Remote workspace reuse is rejected to avoid stale cross-run contamination.
- Slurm monitoring guidance avoids rapid repeated SSH polling.

## Test

- [x] Local backend smoke run completed without DSI remote workspace metadata
- [x] Local Codex continue-AutoResearch round completed successfully
- [x] DSI Slurm smoke testing completed
- [x] DSI Slurm artifacts were copied back and archived during DSI runs
- [x] Continue AutoResearch worked after DSI-backed runs
- [x] Bootstrap and continue-research flows were exercised with the DSI backend

### DIFF
--- a/src/agents/autoresearch_proposer.py
+++ b/src/agents/autoresearch_proposer.py
@@ -19,6 +19,7 @@ import time
 # Add parent directory to path for imports
 sys.path.insert(0, str(Path(__file__).parent.parent))
 
+from core.compute_backend import get_runtime_compute_backend
 from core.security import sanitize_text
 
 
@@ -35,12 +36,81 @@ TRANSCRIPT_FLAGS = {
 }
 
 
+def _skill_root_for_provider(provider: str) -> str:
+    return f".{provider}/skills"
+
+
+def _generate_compute_backend_section(idea_spec: Dict[str, Any], provider: str = "claude") -> str:
+    """Return proposer-only backend constraints for explicit remote backends."""
+    backend = get_runtime_compute_backend(idea_spec)
+    skill_root = _skill_root_for_provider(provider)
+    dsi_skill_path = f"{skill_root}/dsi-slurm/SKILL.md"
+    if backend == "dsi-slurm":
+        return f"""
+═══════════════════════════════════════════════════════════════════════════════
+                              COMPUTE BACKEND: dsi-slurm
+═══════════════════════════════════════════════════════════════════════════════
+
+Runtime execution is pinned to DSI Slurm by `--compute-backend dsi-slurm`.
+Any proposal that requires cluster training, evaluation, or batch execution
+must tell comment mode to use `{skill_root}/dsi-slurm/` by reading
+`{dsi_skill_path}` and following that skill's guidance.
+
+The proposal must preserve this compute invariant: the local workspace is for
+orchestration and reporting only. Comment mode must not run training,
+evaluation, model selection, benchmarking, scored-output generation, smoke
+tests, or result-changing validation locally. Local commands may inspect files,
+edit code, prepare scripts, package inputs, and verify already-copied results
+only. DSI Slurm is the only allowed compute surface for experiment workload.
+
+The proposal should preserve the backend lifecycle contract: setup/discovery
+checks first, use only the runtime-provided remote workspace, cheap smoke job
+when possible, explicit resource requests, and copy-back of all required
+results from the remote workspace to the same relative local paths. Comment
+mode must also copy each terminal job's `dsi-slurm-artifacts/<JOB_ID>/` bundle
+back to the local workspace. NeuriCo runtime creates/removes the remote
+workspace and archives local `dsi-slurm-artifacts/`; comment mode must not
+remove the remote workspace itself.
+
+Do not propose Modal, local GPU fallback, or any other off-machine backend. If
+missing DSI Slurm configuration or access would block the proposed change, make
+that blocker explicit in the proposal rather than suggesting a backend switch.
+
+"""
+    if backend == "modal":
+        return """
+═══════════════════════════════════════════════════════════════════════════════
+                              COMPUTE BUDGET
+═══════════════════════════════════════════════════════════════════════════════
+
+If your proposal would require GPU model training, fine-tuning, or LLM serving
+that exceeds the local container, the workspace may have a compute-backend
+skill available. Do not propose a backend by name. Instead, scope your proposal
+so that:
+
+1. If a compute backend is available, you state the proposal's compute needs
+   (model size, GPU memory, expected wall time) and note that an off-machine
+   backend may be required — the experiment_runner agent will discover and
+   pick the appropriate skill.
+2. If no compute backend is available, propose only changes that fit on the
+   local container (smaller models, fewer steps, eval-only paths).
+
+Treat compute-backend availability as a constraint to scope around, not as a
+licence to propose unbounded training jobs.
+
+═══════════════════════════════════════════════════════════════════════════════
+
+"""
+    return ""
+
+
 def generate_autoresearch_proposal_prompt(
     idea: Dict[str, Any],
     work_dir: Path,
     parent_sha: str,
     attempt_dir: Path,
     templates_dir: Path,
+    provider: str = "claude",
     attempt_history: Optional[list[Dict[str, Any]]] = None,
 ) -> str:
     """
@@ -81,6 +151,7 @@ def generate_autoresearch_proposal_prompt(
         attempt_dir=str(attempt_dir),
         proposal_path=str(attempt_dir / "proposal.md"),
         public_context=context,
+        compute_backend_section=_generate_compute_backend_section(idea_spec, provider=provider),
     )
 
 
@@ -154,6 +225,7 @@ def run_autoresearch_proposer(
         parent_sha=parent_sha,
         attempt_dir=attempt_dir,
         templates_dir=Path(templates_dir),
+        provider=provider,
         attempt_history=attempt_history,
     )
 
--- a/src/agents/comment_handler.py
+++ b/src/agents/comment_handler.py
@@ -140,7 +140,8 @@ def resolve_workspace(
 def generate_comment_prompt(
     idea: Dict[str, Any],
     work_dir: Path,
-    templates_dir: Path
+    templates_dir: Path,
+    provider: str = "claude",
 ) -> str:
     """
     Generate the comment handler prompt by combining template with idea comments.
@@ -156,7 +157,7 @@ def generate_comment_prompt(
     from templates.prompt_generator import PromptGenerator
 
     generator = PromptGenerator(templates_dir)
-    return generator.generate_comment_prompt(idea, work_dir)
+    return generator.generate_comment_prompt(idea, work_dir, provider=provider)
 
 
 def run_comment_handler(
@@ -219,9 +220,34 @@ def run_comment_handler(
     print("-" * 40)
     print()
 
+    from core.dsi_slurm_remote import dsi_slurm_remote_workspace
+
+    with dsi_slurm_remote_workspace(idea, work_dir) as dsi_remote_info:
+        if dsi_remote_info is not None:
+            print(f"DSI remote workspace: {dsi_remote_info['remote_root']}")
+        return _run_comment_handler_with_remote_workspace(
+            idea=idea,
+            work_dir=work_dir,
+            provider=provider,
+            templates_dir=templates_dir,
+            timeout=timeout,
+            full_permissions=full_permissions,
+            dsi_remote_info=dsi_remote_info,
+        )
+
+
+def _run_comment_handler_with_remote_workspace(
+    idea: Dict[str, Any],
+    work_dir: Path,
+    provider: str,
+    templates_dir: Path,
+    timeout: int,
+    full_permissions: bool,
+    dsi_remote_info: Optional[Dict[str, str]],
+) -> Dict[str, Any]:
     # Generate prompt
     print("Generating comment handler prompt...")
-    prompt = generate_comment_prompt(idea, work_dir, templates_dir)
+    prompt = generate_comment_prompt(idea, work_dir, templates_dir, provider=provider)
 
     # Save prompt for reference
     logs_dir = work_dir / "logs"
@@ -267,6 +293,9 @@ def run_comment_handler(
     # Set environment variables
     env = os.environ.copy()
     env['PYTHONUNBUFFERED'] = '1'
+    if dsi_remote_info is not None:
+        env["NEURICO_DSI_REMOTE_ROOT"] = dsi_remote_info["remote_root"]
+        env["NEURICO_DSI_RSYNC_REMOTE_ROOT"] = dsi_remote_info["rsync_remote_root"]
 
     if provider == "gemini":
         env['GEMINI_CLI_IDE_DISABLE'] = '1'
--- a/src/core/agent_runner.py
+++ b/src/core/agent_runner.py
@@ -35,6 +35,7 @@ from datetime import datetime
 # Add parent directory to path for imports
 sys.path.insert(0, str(Path(__file__).parent.parent))
 
+from core.compute_backend import attach_runtime_compute_backend
 from core.security import sanitize_text
 
 
@@ -154,7 +155,8 @@ def _build_agent_command(provider: str, full_permissions: bool = True,
 
 def _run_cli_agent(cmd: str, prompt: str, work_dir: Path,
                    log_file: Path, transcript_file: Path,
-                   tracker: RunTracker) -> Dict[str, Any]:
+                   tracker: RunTracker,
+                   env_extra: Optional[Dict[str, str]] = None) -> Dict[str, Any]:
     """
     Execute a CLI agent with streaming output capture.
 
@@ -162,6 +164,8 @@ def _run_cli_agent(cmd: str, prompt: str, work_dir: Path,
     """
     env = os.environ.copy()
     env['PYTHONUNBUFFERED'] = '1'
+    if env_extra:
+        env.update(env_extra)
 
     # Disable IDE integration for Gemini CLI
     if 'gemini' in cmd:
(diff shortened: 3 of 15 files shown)


AndrewRqy:
Hi! really nice work on the provider-aware skill path discovery and the local-backend gating:) There are a few things I wanted to talk through before we land this:

  One question on scope
  
  src/core/autoresearch.py shows +842 / -31, but reading through it most of that (~830 lines) looks like the bootstrap AutoResearch baseline flow from commit 67cc25e — a separate state machine with its own persistence format that doesn't seem related to the compute-backend selection story. Would it be reasonable to split that out into its own PR? It looks internally coherent and would et each feature get a clean review trail. The actual DSI glue here is just the 10-line _move_dsi_slurm_artifacts_to_attempt hook at :1604, :1691, :1813-1817 — happy to be wrong if those changes are more intertwined than they look from the diff.

  A few cleanup-ordering things I wanted to flag
  
  These are the spots where I'm a little worried about leaking the remote workspace or leaving zombie jobs in the failure paths:

  1. src/core/agent_runner.py:283-297 — create_remote_workspace runs before the try: block that owns remove_remote_workspace. If anything in between raises (transient import, OOM, Ctrl-C during the print), the cleanup finally never fires. The orchestrator path at pipeline_orchestrator.py:503-512 already wraps it the safe way — maybe we could mirror that here?
  2. src/agents/comment_handler.py:244-246 — the finally: removes the remote workspace, but I don't see a scancel before it. If Slurm job is still pending or running when this fires, we'd rm -rf its CWD out from under it. Either a scancel in finally, or some guarantee that the agent has waited on all jobs before exit — happy to discuss which fits the design better.
  3. templates/skills/dsi-slurm/SKILL.md:186-193 — the submit-wait loop uses squeue -h -j ... -o "%T" and [ -z "$STATE" ] as the terminal check. I think this works for single jobs but might misbehave on arrays (one line per task, mixed-state cases). The skill calls out array support at :75-83 and :240-245, so the wait loop probably needs to be array-aware too — or alternatively, we could narrow the skill's array claims. I might be missing something about how arrays land here, though.
  4. src/core/runner.py standalone path — I might be reading this wrong, but I couldn't find a create_remote_workspace call in the non-orchestrator runner code path. With --compute-backend dsi-slurm the prompt section gets emitted (telling the agent to read .neurico/dsi_slurm_remote_workspace.json), but the file wouldn't exist. Is the standalone runner intended to support DSI, or should we treat it as orchestrator-only?
  
  Some smaller things while we're here

  - dsi_slurm_remote.py:27-31 — workspace names containing / or .. would slip past shlex.quote and escape ~/neurico_workspaces/. A simple ^[A-Za-z0-9._-]+$ check would close that.
  - dsi_slurm_remote.py:73-79 — [ -e "$root" ] + mkdir -p could race two pipelines starting near-simultaneously. Bare mkdir "$root" would let the kernel reject the loser atomically.
  - dsi_slurm_remote.py:114-134 — dsi_slurm_remote_workspace is a context manager that I don't see used anywhere; the orchestrator and agent_runner each open-code its logic and have already started to diverge a bit (cleanup-exception handling differs). Worth either adopting it or dropping it.
  - dsi_slurm_artifacts.py:13-36 — fixed .{name}.tmp sibling path; two moves into the same parent would step on each other, and the destination clobber rmtrees any prior attempt's artifacts. tempfile.mkdtemp + per-job-id subdir merge would be safer.
  - compute_backend.py:23-30 — the dual-key write (idea + idea["idea"]) reads cleanly today, but I worry it could drift if anything mutates one and not the other. Picking one source of truth would simplify.
  - No stale-sentinel detection — if a prior run hard-crashed (SIGKILL, OOM) after create_remote_workspace, both the sentinel and remote workspace persist; the next run would crash inside SSH with exit-code-17 and no clear hint of what to do. A startup check with a remediation message would help.
  
  One thing that'd really increase confidence
  
  dsi_slurm_remote.py is the SSH-touching core of this PR (134 lines), and as far as I can tell it doesn't have any direct unit tests. The existing test file covers artifact archival but none of the workspace lifecycle. The run= injectable in the module is set up beautifully for fakes; would it make sense to add a small test_compute_backend.py for the pure functions and a fake-subprocess test for the workspace context manager? That'd let us assert the cleanup-ordering and failure-path claims without needing to live-run against the cluster every time.

  A few smaller doc things (not blocking)
  
  - The SKILL.md doesn't have a "When to use / Prerequisites" section the way modal-training/SKILL.md does — that pattern was really helpful for routing agents.
  - Step 3 (sync files in) doesn't show an rsync example; step 7 does. An agent might land on scp without seeing the pattern.
  - Maybe one sentence on why rapid SSH polling is bad? It's forbidden but not explained, and an LLM that doesn't know the rationale might cut corners.
  - job.json is recorded but I couldn't find a consumer in scoring/ or paper_writer/ — worth either clarifying "this is for
  postmortem only" in the schema doc, or wiring a reader.
  - autoresearch_proposer.txt:36-37 — the PUBLIC CONTEXT header looks like it's missing a top ═══... border that the injected backend section was supposed to close out, so with --compute-backend local you get an orphaned blank line above the header.

Frankbest18:
Thanks Andrew — this review was really helpful. I replaced the branch with a cleaned-up version on top of current `main`, so the PR is now focused on the DSI Slurm/backend-selection work only.

Current diff:
- 15 files changed
- no bootstrap AutoResearch base work carried in this PR
- no local test files included

Addressed:

1. Scope / `autoresearch.py`
   - Cleaned the branch on top of current `main` after the base work landed.
   - The large bootstrap AutoResearch diff is no longer part of this PR.
   - The remaining `autoresearch.py` change is only the small DSI artifact movement hook needed to archive per-attempt Slurm evidence.

2. Remote workspace cleanup ordering
   - `agent_runner.py` now uses the shared `dsi_slurm_remote_workspace(...)` context manager.
   - Comment mode is also wrapped in the same lifecycle.
   - This removes the create-before-try/finally gap and keeps the cleanup path consistent with the orchestrator path.

3. Standalone runner support
   - Standalone agent execution now supports `--compute-backend dsi-slurm`.
   - It creates the runtime remote workspace metadata before launching the agent, so the prompt no longer points to a missing `.neurico/dsi_slurm_remote_workspace.json`.

4. Workspace name safety
   - Added conservative workspace-name validation: letters, numbers, `.`, `_`, and `-` only.
   - This prevents path escape through names like `/` or `..`.

5. Stale remote workspace handling
   - Remote workspace creation now removes any stale deterministic remote root first, verifies it is gone, then creates a fresh workspace.
   - Cleanup also verifies that the remote root is gone after `rm -rf`.
   - If cleanup fails, the warning is explicit instead of silently assuming success.

6. Context manager divergence
   - The `dsi_slurm_remote_workspace(...)` context manager is now the shared path used by experiment/comment/standalone execution instead of unused dead code.

7. Artifact move safety
   - DSI artifact archival now uses a unique temp directory.
   - It merges per-job artifact directories into the destination instead of clobbering the whole destination, so multiple jobs/attempts are safer.

8. Compute backend runtime marker
   - I kept the dual in-memory marker intentionally because runtime paths sometimes pass the full idea object while prompt generation often receives `idea["idea"]`.
   - To avoid persistence drift, the marker is stripped before writing idea metadata via `without_runtime_compute_backend(...)`.
   - So this remains a compatibility bridge, not persisted idea state.

9. Slurm skill cleanup
   - Added “When to Use” and “Prerequisites”.
   - Added an rsync example in the sync step.
   - Added rationale for avoiding rapid SSH polling.
   - Clarified that `job.json` is audit/postmortem metadata, not a scoring input.
   - Updated the wait-loop template to collect unique states from `squeue`, which is safer for mixed-state / array-like output.

10. Prompt/template cleanup
   - Fixed the local-backend behavior so backend guidance is not injected for local runs.
   - The proposer/comment/experiment prompts now receive backend-specific guidance only when a non-local backend is selected.

Not fully addressed / design choice:

1. Runtime `scancel` in `finally`
   - I did not add a blind runtime-level `scancel` in `finally`.
   - The runtime currently does not own a reliable structured registry of Slurm job IDs submitted by the agent.
   - The current design keeps submission, waiting, cancellation, and copy-back as agent responsibilities because the agent knows which jobs it created and which outputs matter.
   - The skill contract now makes this stricter.
   - If we want runtime-owned cancellation later, I think the right next step is a real job registry written during submission, rather than best-effort cancellation from incomplete state.

2. Tests
   - I wrote focused local tests while iterating on `dsi_slurm_remote.py` and artifact archival, but did not include them in this PR because we decided to keep tests local for now.
   - Happy to add them in a follow-up if we want this covered in CI.

Validation after the cleanup:

- Local backend smoke run: no DSI remote workspace metadata and no Slurm/Modal guidance injected.
- DSI Slurm tiny AutoResearch smoke run completed.
- Base Slurm job completed and archived under `logs/dsi-slurm-artifacts/`.
- AutoResearch Slurm job completed and archived under `logs/experiment-autoresearch/<parent_sha>/attempt_*/dsi-slurm-artifacts/`.
- Root transient `dsi-slurm-artifacts/` was removed after archive.
- Remote workspace cleanup was verified after the run.

AndrewRqy:
Great work! I merged pr #122 and now PR #132 conflicts on src/core/runner.py. Could you rebase dsi-slurm onto current origin/main and resolve? Three of the four are trivially additive. The fourth is in the new if bootstrap_autoresearch_baseline: block: its self._copy_workspace_resources(work_dir) call needs to become self._copy_workspace_resources(work_dir, compute_backend=compute_backend) to match the signature PR 132 introduced.

Frankbest18:
Thanks for flagging this. PR #132 has now been rebased onto current `main` after PR #122 landed.

The conflict was only in `src/core/runner.py`. I resolved it by keeping both sides:

- kept PR #122’s `bootstrap_autoresearch_baseline` and `proposer_timeout` arguments
- kept PR #132’s `compute_backend` argument
- kept PR #122’s stricter three-way AutoResearch mode validation
- kept PR #132’s backend normalization / runtime attachment
- updated the new bootstrap AutoResearch baseline path so workspace resource copying remains backend-aware:

```python
prepare_workspace=lambda bootstrap_work_dir: self._copy_workspace_resources(
    bootstrap_work_dir,
    compute_backend=compute_backend,
)
```

AndrewRqy (approved):

PR #133, opened 2026-06-24
modal-skills: chain reliability, vllm correctness, docs reconcile

AndrewRqy:
## Summary

Bug-fix follow-up to the modal-training / modal-vllm skills landed in #124. Each item was surfaced by post-merge code review or live-run validation; commit messages have the per-change rationale.

## What's fixed

- **Chained-stage pulls** — eval re-uploads `val.jsonl` + `artifacts/final/` (was FileNotFound on the dataloader + silent base-model fallback); `register()` REPLACES `pull_manifest` instead of merging across stages with cascade-deleted volumes; `data_prep` + `eval` emit `*_config.json` provenance the recipes doc requires
- **Sentinel safety** — atomic + `fcntl`-locked writes (concurrent stages couldn't tear or lose updates); cross-experiment re-key refuses with the prior `exp_id` + recovery CLI in the error; `upload_to_volume` refuses on torn-down state; directory pulls stage to `<dest>.prev` and restore on failure
- **modal-vllm** — scripts find the sibling `modal-training` by walking up from `__file__` (was relative-to-cwd, broke for non-cwd invocations and codex/gemini providers); endpoint URL derived via `modal.Function.from_name().get_web_url()` (modal CLI 1.4's `app list --json` has no URL field; synthesis was stale for env-scoped apps); `capture-endpoint` no longer re-runs `register()` (was clobbering `pull_complete` on every invocation); recursive secret redaction
- **Sweep dispatch** — orchestrator reads the sentinel's `endpoint_captured`/`apps` and picks `modal-vllm`'s sweep when applicable; without this a vllm-only workspace's crash-cleanup left live proxy-auth tokens in `artifacts/`
- **Scaffolders** — `--force` overwrite guard on both (was silent clobber); `resolve_secrets` deduped across both
- **Docs** — vllm proxy-auth flow matches code (dashboard mint + env-var pass, not the never-existed `modal token new --proxy`); volume-root vs container-mount paths documented next to `PULL_MANIFEST`; GPU prices reconciled between `SKILL.md` and `modal_basics.md`; `status` CLI signature corrected (no `--exp-id`); `compute-backend` discovery gets a concrete `grep`

## Validation

Live-tested end-to-end against Modal:

- prep → lora-train → eval chain on Qwen2.5-0.5B (L4, 5 steps): all artifacts pulled, adapter correctly loaded in eval, clean leave-no-trace teardown
- vLLM deploy + capture-endpoint: SDK URL matched the real deploy URL, sentinel `pull_complete` preserved across re-invocations
- B4 cross-experiment guard: refused with the prior `exp_id` named in the error
- modal-vllm sweep: correctly dispatched, live endpoint JSON cleared on teardown
- No modal envs / apps / secrets / volumes left after the sweep

## Test plan

- [x] AST + structural smoke checks across all 4 templates (16 assertions)
- [x] Sentinel concurrency harness — 8 writers × 50 ops, all 400 land, no torn writes
- [x] Mocked-`_run` regression tests for `pull_all` unknown-volume, `upload_to_volume` torn-down, directory pull rmtree-restore (26 assertions)
- [x] modal-vllm doctrines: non-cwd import, `--force` guard, recursive redact, sweep dispatch, URL discovery (26 assertions)
- [x] Docs reconcile checks against code claims (21 assertions)
- [x] Live: prep + lora-train + eval chain end-to-end on Modal
- [x] Live: vllm deploy + capture-endpoint + sweep + leave-no-trace verification


Frankbest18 (changes requested):
Thanks for the PR. I only noticed a few issues worth further attention.

### Cross-experiment guard is not atomic

`register()` checks the existing sentinel before acquiring the sentinel lock. It then creates the Modal environment and only later writes the sentinel through `update_sentinel()`.

That leaves a race where two concurrent registrations with different `exp_id`s can both pass the pre-check before either writes the sentinel. The later write can overwrite the earlier local record, leaving the earlier Modal environment without a local recovery/teardown pointer.

---

### `pull_all()` can mark the wrong registration complete

`pull_all()` loads the sentinel and `pull_manifest` once, performs Modal pulls, and then writes `pull_complete` back into whatever sentinel exists at the end. Because chained stages reuse the same `exp_id` while replacing `pull_manifest`, `exp_id` alone is not enough to prove the sentinel still represents the same stage that was pulled.

A slow pull from an earlier registration could finish after a newer registration has replaced the manifest, then incorrectly set `pull_complete=True` for the newer stage.

---

### Generated vLLM capture still depends on cwd

The PR fixes non-cwd sibling discovery inside `modal-vllm/scripts/lifecycle.py`, but the generated vLLM serve template still imports lifecycle through `Path(".claude/skills/modal-vllm/scripts")`. That path is relative to the current working directory, not the generated script.

So `python src/modal_serve.py capture-endpoint` works from the workspace root, but `python /path/to/workspace/src/modal_serve.py capture-endpoint` from another directory can leave `lifecycle = None` and later crash at `lifecycle.capture_endpoint(...)`.

---

### `capture_endpoint()` can overwrite concurrent sentinel updates

`modal-vllm/scripts/lifecycle.py` uses `base.load_sentinel()` + local mutation + `base.save_sentinel()` to set `endpoint_captured=True`. `save_sentinel()` makes the file write locked/atomic, but the full read-modify-write is not locked. That means another process can update the sentinel between the read and the write, and `capture_endpoint()` can then write back an older copy of the sentinel with only `endpoint_captured` added.

For example, a concurrent `pull_all()` could mark `pull_complete=True`, then `capture_endpoint()` could overwrite that with its stale copy and set `pull_complete` back to the older value.

AndrewRqy:
Thanks a lot for the review! I check the 4 problems and try to fix them in this new commit, let me know if they looks good rn :)

Frankbest18 (approved):

PR #142, opened 2026-07-22
Local idea submission with enforced local resources and evaluation contracts

AndrewRqy:
## What this adds

Users running NeuriCo locally may have datasets and evaluation functions already on disk, but there was no way to declare them when submitting an idea. This PR adds a local submission path and makes those declarations binding through the pipeline.

## Changes

- `./neurico submit-local <idea.md>` converts a local markdown or text file to idea YAML, with the same convert / save / `--submit` / `--run` flow as `fetch`. Local paths mentioned in the file must survive conversion verbatim; the CLI fails if any are dropped.
- Schema v1.2 adds `idea.local_resources` (datasets and functions with path, usage, entrypoint, optional `required_for_evaluation`) and `idea.evaluation` (metric names, targets, results format). Both are validated at submit time.
- Before any agent runs, declared files are copied into the workspace (`datasets/local/`, `code/local/`) and idea paths are rewritten workspace-relative. A missing path fails the run. Staged functions get a sha256 fingerprint.
- The resource finder, experiment runner, and session instructions render the staged resources with their usage. The resource finder is instructed to verify them and not download substitutes.
- The rule maker transcribes user metrics into targets.json verbatim with `"source": "user"` and must call mandated functions from eval.py. A new eval_verifier agent reviews scoring/ against the declarations before sealing; on failure the rule maker reruns once with the findings appended. The scorer refuses to run if a fingerprinted function changed after staging.
- Parity changes outside the new feature: `fetch_from_ideahub.py` now prefers OpenRouter keys for conversion (matching #139), and both converters drop the `author: Unknown` placeholder.

## Testing

- 40 new unit tests (validation, path detection, staging, prompt rendering, verifier plumbing, integrity guard, sealing). Full suite passes at 106.
- Live conversions: local paths land in `local_resources` verbatim; a mixed idea separated a HuggingFace dataset (background) from local paths (local_resources) and a mandated scorer from a helper function.
- Verifier tested standalone: passed a compliant harness, rejected a sabotaged one (reimplemented metric, renamed metric, lowered threshold), each violation with quoted evidence.
- Full pipeline replication on a toy idea: staging ran, the resource finder verified the staged files and downloaded no substitutes, the rule maker passed verification on the first attempt, the runner reached test_accuracy 0.925 against the user target 0.85, and the scorer succeeded. Modifying the staged function afterward made the scorer refuse; restoring it scored again.


Frankbest18 (changes requested):
Thanks for the greate work, here are some feedback that might need more attention:

### P1 — `submit-local` does not reliably carry local files into Docker

The advertised Docker flow has three related failures:

- `./neurico submit-local idea.md` does not mount an ordinary relative host idea file into the container.
- `--submit --run` invokes the runner from the submission container and bypasses the host-side `./neurico run <idea-id>` path that adds mounts from `ideas/mounts/<idea-id>.txt`.
- Relative `local_resources` paths are accepted/documented but skipped when mounts are collected.

As a result, a user can submit an idea declaring a local dataset or function, then have the run fail because the container cannot see the declared source.

### P1 — staging destinations are neither contained nor unique

Dataset `name` is used directly in the destination path. A value such as `../../outside` or an absolute path can escape `datasets/local`. Duplicate dataset names merge directories, and two functions with the same basename overwrite one another in `code/local`.

### P1 — conversion can report path “faithfulness” without creating a usable local resource

The faithfulness check only verifies that a path appears somewhere in the converted idea. In the no-API-key fallback, the full input remains in `background.description`, so a required local path can pass this check while never entering `local_resources`. It then is neither mounted nor staged.

### P1 — evaluator-function integrity check fails open

Expected hashes are stored in worker-visible `.neurico/idea.yaml`. If that file is deleted, has no `local_resources`, or an entry has no hash, `staged_function_mismatches()` returns no mismatch and scoring continues.

For a feature intended to enforce a user-mandated evaluator, the expected integrity metadata is itself mutable and missing metadata is treated as success.

### P2 — scoring-only evaluator requirements appear in ordinary prompts

`required_for_evaluation` is rendered as “MANDATORY FOR EVALUATION” in normal research prompts, even when scoring is disabled. NeuriCo’s ordinary research flow is intentionally unscored, so this binding requirement changes behavior outside the scoring pipeline.

### P2 — mount construction uses `eval`

Host mount arguments are assembled into a shell string and reparsed with `eval`. This is brittle for ordinary filenames containing spaces or shell-significant characters, and it unnecessarily treats path data as shell syntax.

### P2 — verifier enforcement is not robust enough

The verifier receives write access to the live scoring files it is meant to review, and runtime does not verify that those files remain unchanged.

Also, the verdict uses `bool(verdict.get("pass"))`; valid JSON such as `{"pass": "false"}` is therefore accepted as a pass.

### P2 — target-less metrics conflict with verifier requirements

The schema permits a metric with no `target` and says the rule maker should derive one with `source: derived`. The verifier, however, requires each user-declared metric to carry a target with `source: user`.

A valid submission can therefore become impossible to verify correctly.

### P2 — staging is not proven idempotent across real continuations

The idempotence test reuses the mutated in-memory idea. A real continuation reloads the original submitted idea, which can cause resources to be copied again, require the original host source again, and merge stale data into a prior staged directory.

### P2 — verifier timeout is ineffective while stdout remains open

The process output is read until EOF before `wait(timeout=...)` is called. A stuck verifier that keeps stdout open can prevent the code from ever reaching the timeout check.

Lower-priority follow-ups:

- Retry logs and verdicts overwrite the first failed attempt, weakening auditability.
- Absolute host `source_path` values are stored in the GitHub-backed workspace contract, exposing host-machine metadata.

Frankbest18:
Hi Andrew! Thanks for the detailed feedback for PR146! Could you submit them as followups of PR146 so they can be better organized and documented? Thanks!

AndrewRqy:
oh damn i submit them under the wrong pr lol, i will fix this right now

AndrewRqy:
Thanks for the thorough review, every item is addressed in the latest commit. Point by point:

Local files not carried into Docker. cmd_submit_local now resolves the idea file to an absolute host path (with an existence check) and mounts its directory at /input:ro for relative and absolute paths alike. --run is intercepted on the host: after a successful submit the idea id is extracted and dispatched to cmd_run, the one path that applies the ideas/mounts/<id>.txt mounts (and submit_local.py refuses an in-process run when it detects it is inside Docker). Relative local_resources are canonicalized to host-absolute at submit time (canonicalize_local_paths, with the host cwd mounted at its identical path and passed as NEURICO_HOST_CWD), and collect_host_paths now warns loudly instead of silently skipping.

Staging containment / uniqueness. New _staging_destination(): names are reduced to basenames (traversal and absolute names cannot escape the staging dirs), duplicates get a deterministic numeric suffix instead of silently merging or overwriting, plus a containment assertion. Covered by new tests.

Faithfulness false-positives. missing_paths_in_idea is now structural: a path must land in local_resources or background.papers[].path (the locations staging and mounting actually read); surviving only in prose fails the gate. The no-API-key fallback warns explicitly that detected paths cannot be auto-extracted.

Integrity fails open. staged_function_mismatches now takes the trusted submitted idea (held by the orchestrator, outside the workspace) and fails closed. The verified path is derived from the contract, not the worker-writable record, and expected bytes come from the strongest available reference: the read-only source file (mounted during Docker runs), then the contract's recorded sha, then the workspace record as a last resort; no reference at all is a mismatch. Forged workspace hashes and record redirects at decoy files are both defeated (regression tests included). Threaded through run_scorer(idea=) at every call site.

Scoring-only requirements in ordinary prompts. scoring_enabled is threaded through prompt and session generation, so required_for_evaluation renders only in scored runs. Tests assert both modes.

eval for mount construction. cmd_run and cmd_submit_local are rebuilt as argv arrays; sidecar paths are literal array elements, never shell-parsed. (The pre-existing eval strings in cmd_submit and a few other commands carry only trusted/static args; happy to convert them in a follow-up.)

Verifier robustness. The reviewed files (eval.py, targets.json, rule_maker_log.md) are snapshotted before the agent runs; any modification, including creating an absent file, is restored and converted into a failing read_only verdict. Verdict parsing is strict (pass is True via interpret_verdict); {"pass": "false"} now fails with an explicit violation.

Verifier timeout. stdout is drained on a helper thread and process.wait(timeout=...) is enforced on wall clock; on timeout the process is killed and any partial verdict is discarded (fail closed).

Target-less metrics. Aligned end-to-end on derived targets: the schema documents the optional target; the rule maker derives one tagged source: derived (documented in rule_maker_log.md) and never fabricates source: user; the verifier accepts user exactly when the user stated a target and derived exactly when they did not, and fails fabricated provenance.

Idempotence across real continuations. Staging is redesigned around contract reloads: already-staged detection by path prefix, existing destinations kept (never merged over), functions refreshed from source on reload so the recorded sha matches staged bytes, and a staged copy survives a missing source (cross-machine continuation). The test reloads a pristine original spec and asserts no stale-merge via a drift marker.

Follow-ups. Retry auditability: verifier and rule-maker artifacts are now per-attempt (…_attempt{n}…), and a superseded verification.json is archived rather than deleted. Host-path exposure: the workspace .neurico/idea.yaml is written with source_path and metadata.source_path redacted (both writers); absolute paths remain only in the local ideas/ dir and the gitignored mounts sidecar.

Frankbest18 (changes requested):
Thanks for the thorough follow-up. I rechecked the latest commit and the major issues appear addressed with three remaining minor issues:

### P2 — `required_for_evaluation` is still mandatory in the normal resource-finder path

The main research prompt and session instructions now correctly gate this requirement on `scoring_enabled`. However, `generate_resource_finder_prompt()` has no scoring-mode argument and still unconditionally renders:

```text
MANDATORY: all evaluation must run through this function
```

when `required_for_evaluation` is true. The accompanying test also expects that text in the normal resource-finder prompt.

Resource finding runs in NeuriCo’s ordinary unscored workflow, so this still applies a scoring-only obligation outside scoring mode.

### P2 — verifier verdict validation is not yet strict

The new `interpret_verdict()` correctly rejects non-Boolean values such as:

```json
{"pass": "false"}
```

However, this incomplete verdict is still accepted:

```json
{"pass": true}
```

Runtime does not require or validate `checks`, validate the allowed check values, validate `violations`, or ensure the aggregate `pass` value is consistent with those checks. The verifier template requires this structure, but the runtime does not enforce it.

### P2 — structural path faithfulness excludes documented background locations

The conversion instructions allow a local path with no declared usage to remain in:

- `background.datasets[].source`, or
- `background.code_references[].repo`.

But `_usable_path_strings()` only recognizes `local_resources` entries and `background.papers[].path`. A conversion that follows the documented background-dataset or code-reference behavior can therefore be rejected as having dropped the path.

AndrewRqy:
Thanks a lot for the followup, the issue should be addressed in this commit, i also do some revision on styling, feel free to ignore them

Frankbest18 (changes requested):
Thanks for the latest follow-up. I rechecked the new implementation. The resource-finder scoring gate and the documented background-path locations are now addressed.

One verifier-schema issue remains:

### P2 — a passing verifier verdict can omit required checks

`interpret_verdict()` now correctly rejects a missing `checks` object, invalid check values, malformed violations, and non-Boolean values such as:

```json
{"pass": "false"}
```

However, it only requires `checks` to be a non-empty mapping. It does not require all three check names mandated by `templates/agents/eval_verifier.txt`:

```json
"checks": {
  "routing": "pass" | "fail" | "not_applicable",
  "transcription": "pass" | "fail" | "not_applicable",
  "format": "pass" | "fail" | "not_applicable"
}
```

Therefore, this incomplete verdict is accepted as a pass:

```json
{
  "pass": true,
  "checks": {
    "routing": "pass"
  },
  "violations": []
}
```

That verdict never reports whether the user’s metrics/targets were faithfully transcribed or whether the requested result format was honored. Since `not_applicable` is already part of the verdict vocabulary, a valid verdict should still explicitly report every check; otherwise a verifier can silently skip part of the evaluation contract while the pipeline accepts the result.

AndrewRqy:
Thanks a lot for the followup, the lastest commit should address this issue

Frankbest18:
Thanks for the latest updates. I rechecked the ordinary non-HITL workflow specifically and only two minor verifier robustness issues remain:

### P2 — an all-`not_applicable` verdict is accepted even though a contract exists

The evaluation verifier only runs when `has_user_eval_contract()` finds at least one declared contract component:

- a mandated evaluation function,
- an evaluation metric, or
- a requested results format.

Therefore, whenever the verifier runs, at least one of `routing`, `transcription`, or `format` must be applicable.

However, `interpret_verdict()` accepts this as a successful verdict:

```json
{
  "pass": true,
  "checks": {
    "routing": "not_applicable",
    "transcription": "not_applicable",
    "format": "not_applicable"
  },
  "violations": []
}
```

This returns a pass even though it silently evaluates none of the contract that caused the verifier to run. The implementation validates the presence and vocabulary of the check fields, but it does not validate their applicability against the submitted idea.

### P2 — a malformed `violations` container crashes verdict processing

`interpret_verdict()` currently begins with:

```python
violations = list(verdict.get("violations") or [])
```

It later validates individual violation entries, but it does not first validate that `violations` is an array. For example:

```json
{
  "pass": true,
  "checks": {
    "routing": "pass",
    "transcription": "pass",
    "format": "pass"
  },
  "violations": 1
}
```

raises:

```text
TypeError: 'int' object is not iterable
```

This exception escapes normal verdict rejection. Instead of treating the malformed verifier response as a failed verification and following the intended retry path, the ordinary rule-maker stage terminates with an exception.

AndrewRqy:
Thanks a lot for the followup! Both issue should be addressed in this new commit:)

Applicability: verdict validation is now driven by one declarative table, VERDICT_CHECKS, mapping each check to the contract component that makes it applicable (the template's "only when..." rules). has_user_eval_contract, the mandated check list, and the new applicability validation all derive from it, and interpret_verdict(verdict, contract) rejects any check the contract makes applicable that is reported not_applicable. So the all-not_applicable verdict fails, and so does waving off even a single applicable check.

Malformed violations: the container is validated once at the top; any non-list value (int, string, mapping) is recorded as a violation and treated as empty, and every later read uses the validated list. Malformed verifier output now takes the normal retry path instead of raising. Both behaviors are unit-tested (all-not_applicable rejection, applicability flipping when results_format is added, and never-raises cases for scalar/string/dict containers).

The commit also closes some siblings of your earlier findings: the staged-function integrity check now requires the trusted idea (the legacy fail-open mode is deleted; two HITL scorer call sites were affected, happy to split that out if you prefer), rule_maker's ordinary path had the same stdout-EOF-before-timeout hang you found in eval_verifier and now uses the shared deadline-aware runner, and eval_verifier's duplicated launch infra was converged onto core/agent_cli + run_prebuilt_cli_agent. Planned as a follow-up PR to keep this one reviewable: deduplicating submit_local.py against fetch_from_ideahub.py and the run.sh flag handling.

Frankbest18:
There is one remaining ordinary verifier issue in the latest update.

**[P2] A verdict can pass while simultaneously reporting a violation**

`scoring/verification.json` is written entirely by the eval-verifier agent, so its fields are not guaranteed to be internally consistent. The current validator accepts a verdict such as:

```json
{
  "pass": true,
  "checks": {
    "routing": "not_applicable",
    "transcription": "pass",
    "format": "not_applicable"
  },
  "violations": [
    {
      "check": "transcription",
      "detail": "target was copied incorrectly",
      "evidence": "targets.json"
    }
  ]
}
```

The implementation rejects `"pass": true` when a check is marked `"fail"`, but it does not reject `"pass": true` when the agent supplies a nonempty, well-formed `violations` list. As a result, the pipeline can accept the scoring contract without retrying the rule maker even though the verifier explicitly reported a defect.

There is also a related, lower-impact schema gap: `"violations": null` and a missing `violations` field are treated as an empty array and accepted in a passing verdict, despite the agent’s required output format specifying that `violations` is an array.

AndrewRqy:
Both items fixed: pass: true with any recorded violation is now rejected (mirror of the existing pass=false-with-empty-list rule), and violations is required to be an always-present array, so null/missing fails instead of normalizing to empty.
The invariant is now stated in the code: a verdict passes iff pass is true, all three checks are reported, every applicable check is pass, and violations is exactly empty.

I also reinforce the prompt: the template now orders the output checks → violations → summary → pass and instructs the verifier to decide pass last as a conclusion from the written evidence, with an explicit "no pass with reservations" rule so the model is steered away from writing the inconsistency, and the runtime catches it if it happens anyway.

Frankbest18 (changes requested):
Thanks, Andrew. I rechecked the latest update and confirmed that the remaining verifier issues are resolved.

Could you please update the branch and resolve the merge conflicts? Once that is done, I’ll perform a final check of the resolved code before approving.

AndrewRqy:
Thanks for the headsup, the conflicts should be gone now

Frankbest18 (approved):

PR #143, opened 2026-07-23
Continue-research mode: adopt an existing repo and optimize it under user constraints

AndrewRqy:
Stacked on #142: only commit 68d1310 is new here. The diff collapses to that commit once #142 merges.

## What this adds

NeuriCo can now continue research from any existing repository. The user provides a repo (local path or GitHub URL) and an intention file stating what to optimize and what must not break; NeuriCo adopts the repo, scores its current state as an AutoResearch baseline, and iterates toward the goal.

## Changes

- `./neurico continue-research <repo> <intention.md>` converts the intention into an idea with a new `continuation` schema section: `source_repo`, `goal`, and `invariants` of three kinds (`protected_path`, `check`, `statement`). Explicit thresholds in the goal also transcribe into `evaluation.metrics`.
- Adoption (`src/core/repo_adoption.py`) clones or copies the repo into a workspace, never in place, ensures a git anchor for checkpoints, and optionally creates a NeuriCo GitHub backup repo under the same `--no-github` semantics as normal runs (the source remote is kept as `upstream`).
- The runner branches on `idea.continuation`: adopt, stage declared local resources, build the baseline through the existing bootstrap rule maker path, then run `continue_from_current_best`. A workspace with an AutoResearch checkpoint resumes instead of re-adopting.
- Invariants are enforced by kind. `check` commands become guardrail properties in targets.json with pass targets, so the comparator's no-lost-satisfied-properties rule rejects any iteration that breaks them. `protected_path` gets a deterministic git-diff guard that rejects the iteration before scoring. `statement` invariants are rendered as binding instructions and reviewed by the verifier.
- The eval_verifier now also gates the bootstrap rule maker (one retry with findings appended), and checks guardrail transcription.
- Datasets declared `sealed: true` stage into `data/.test/` and inherit the existing sealed-groundtruth handling: hidden from agents, readable only by eval.py.
- Host paths an idea depends on are written to a bash-parseable sidecar (`ideas/mounts/<idea_id>.txt`) at submit time; `cmd_run` mounts each one read-only at its identical in-container path, so local repos and datasets work in Docker mode.
- The proposer prompt carries the continuation goal, and comment-mode prompts carry a binding invariants banner.
- Refactor: the converter CLIs (`submit_local.py`, `fetch_from_ideahub.py`, new `continue_research.py`) share their conversion helpers through `src/cli/idea_conversion.py` instead of duplicating them.

## Testing

- 30 new unit tests in `tests/test_continuation.py` plus sidecar tests; full suite passes at 141.
- Live conversion: a real intention file produced all three invariant kinds correctly (verbatim check command, protected path, prose statement), a sealed benchmark decided by the converter from the stated usage, a mandated evaluation function, and the goal thresholds as metrics.
- Full live replication on an adopted toy repo: baseline scored the user metric at 0.76 with the test-suite guardrail at 1.0; one AutoResearch iteration fitted the decision threshold on the training data only and reached 0.94; the comparator accepted with a strict-superset verdict; the protected path was untouched and the sealed benchmark never left `data/.test/`.
- Sealed staging, protected-path guard (modified, new, and untouched files), adoption idempotency, and in-repo relative resources are unit-tested; the verifier retry and the live protected-path rejection ran only in unit tests since the live agents complied on first attempt.


Frankbest18 (changes requested):
Thanks for the work on this, Andrew. The overall direction makes sense. There are a few behaviors that I think need attention before merging.

### P1 — The combined command can still publish without effective GitHub controls

The `continue-research --submit --run` path does not appear to expose or preserve controls such as `--no-github` and `--private` when it hands the submitted idea to the normal run command.

As a result, if GitHub integration is configured, someone continuing a local repository may not have an effective way to prevent publication or require that the resulting repository be private. This is especially risky here because the source may be an existing local repository containing unpublished work.

### P1 — Sealed resources are not consistently hidden from the research agent

I found several paths through which a resource declared with `visibility: sealed` can remain accessible.

First, `_planned_destination()` initially chooses the sealed destination, but it returns the original path when the resource is a relative path that already exists inside the adopted repository. In that case, the resource remains in its ordinary repository location where the research agent can read it.

Second, `data/.test` does not appear to be included in the list of paths hidden during bootstrap, so bootstrap-stage tooling may be able to inspect held-out data.

Finally, the original host resource remains mounted in the research container. This means that even if the staged copy is placed in a sealed location, the same underlying source may still be accessible through the host mount.

### P1 — Check commands can modify protected files after the protection check

The protected-path guard runs before scoring, but check commands execute afterward in the live, writable research workspace. The workspace is then checkpointed without another protected-path check.

This creates a window where the protected-path guard has already passed but a check command can still modify a protected file. If that happens, the modification can become part of the accepted checkpoint.

I would say this does not necessarily require a malicious command. Test and evaluation commands can have side effects, such as updating snapshots, generated files, caches, or configuration. The issue is that protected state can change after it has been validated without that change being detected.

### P1 — Local adoption can copy files from outside the source repository through symlinks

When a local repository is adopted, symlinks are dereferenced instead of preserved. If the source repository contains a symlink to something outside the repository, the target’s contents can be copied into the NeuriCo workspace as a normal file or directory.

That copied content can then be included in the workspace’s Git history and potentially published. For example, a symlink to a private dataset, local configuration file, or other external material could unintentionally turn that external content into repository content. It can also change the behavior of legitimate symlinks or unexpectedly copy a very large external directory.

### P1 — Some valid protected paths are normalized incorrectly or are invisible to detection

Protected paths are currently normalized with `lstrip("./")`. This removes any combination of leading `.` and `/` characters rather than the specific `"./"` prefix.

For example:

- `.github/workflows` becomes `github/workflows`
- `.env` becomes `env`
- `../src` becomes `src`
- `.` becomes an empty path

The declaration may therefore pass validation while the runtime guard watches a different path from the one the user provided.

There is also a separate issue with ignored files. The guard relies on Git’s modified and untracked file reporting, but ignored files are not included there. Protected ignored files—such as model weights, datasets, generated configuration, or local environment files—can therefore change without the guard detecting the modification.

Together, these cases mean that a user can declare a protected path but not receive the protection they reasonably expect.

### P1 — A failing check invariant can become the accepted baseline

A check invariant represents something that must keep passing, but baseline construction appears to accept the baseline when the scorer completes successfully even if the check itself reports failure.

Once that failing baseline has been accepted, the comparator only prevents the loss of properties that were previously satisfied. If the check failed for both the baseline and a later candidate, the candidate has not technically changed a satisfied property into an unsatisfied one. It may therefore still be accepted because another metric improved.

### P2 — Fallback conversion can run without the structured protections from the intention

If the LLM conversion is unavailable or fails, the template fallback retains the original description as general prose, but it does not extract structured invariants or evaluation metrics.

This means the user’s words may still appear in the generated idea, but NeuriCo no longer has the structured fields needed to enforce them. Numeric thresholds are not included in the scoring contract, and protected-path or check requirements are not registered as runtime invariants.

The converter warns that these fields were not extracted, but `--submit --run` can still continue automatically. The generated idea passes schema validation because those fields are optional. There is a hard check for local paths dropped during conversion, but no corresponding check for non-path constraints.

The result is that NeuriCo can know that the conversion is incomplete and still begin the run without protections contained in the original intention.

### P2 — `--force-fresh` can continue from the previous result

At the runner level, `--force-fresh` bypasses the normal resume branch. However, it continues using the same workspace, and the lower-level adoption and bootstrap functions have their own reuse behavior.

`adopt_repository()` sees the existing adoption record and leaves the workspace in place. Baseline construction then sees the existing AutoResearch current-best checkpoint and returns it instead of creating a new baseline.

Therefore, a run started with `--force-fresh` can still continue from the previous current-best checkpoint. It may not read the source repository again or establish a new baseline.

### Non-blocking concern — Statement invariants are not verified

I also wanted to flag one design concern, although I do not see it as the same kind of deterministic bug as the issues above.

A `statement` invariant is provided to the research agent as a natural-language instruction, but it is not part of the mechanically verified contract. A candidate can violate it and still be accepted if the executable checks and metrics pass.

I understand that arbitrary natural-language constraints cannot always be verified automatically. My concern is mainly that the current language can give the impression that every invariant is enforced as a hard guardrail, when statement invariants actually depend on the agent correctly interpreting and following the prompt.

Could you please take another look at the issues above? Thanks for the great work!

AndrewRqy:
thanks a lot for the feedback! after some rethinking of the architecture i redesign the continue research mechanism to keep sealed data out of the agent's reach by construction instead of hiding it after the fact. Two structural changes, then your points one by one:

Sealed store. Sealed datasets no longer stage into the workspace at all. They live in a sealed store sibling (<workspaces>/.sealed_store/<name>/, same convention as .scoring_sealed), and data/.test is now a materialization, not a location: the store is copied into the tree being scored only while eval.py runs, and removed after. eval.py's contract is unchanged. Since the workspace never contains sealed bytes, the manifest trimmer, rule maker, and experiment runner cannot read them at any stage, with no seal/unseal choreography needed.

Isolated scoring for iterations. Continuation iterations now score a frozen candidate through your run_isolated_scorer from the HITL work: the closure seals the scoring files, commits the candidate to a checkpoint, checks that sha out into a detached worktree, materializes the sealed store there, and brings back only the results.json review copy. This turned out to be exactly the right machinery, so thanks for having built it. It means check commands physically cannot modify the live workspace, which resolves your timing concern more strongly than a re-check would.

On the specific points:

1. GitHub publication: --no-github/--private/--public are now plumbed through continue-research to the dispatched run, and a local source repo defaults to a private backup with a printed notice; --public is the explicit opt-in. Remote URLs keep the ordinary default.
2. Sealed visibility: covered by the store (above). The remaining host-mount exposure is handled by two-phase dispatch: sealed sources, any declared ancestor path containing one, and the source repo itself mount only into a short preparation container that stages and exits; the research container never mounts them.
3. Check commands vs protected paths: structurally closed by isolated scoring; scoring runs in a throwaway worktree. The protected-path guard itself is also rebuilt on filesystem content fingerprints, so git-ignored files are covered.
4. Symlinks: adoption now copies with symlinks=True and warns in both directions (targets outside the repo, and absolute targets into the repo that will dangle once unmounted).
5. Normalization + ignored files: the lstrip bug is fixed with component-wise canonicalization (.github/.env keep their dots, any .. component is rejected, including trailing src/..), and the fingerprint guard sees ignored files by construction. Invalid declarations are now submit-time errors rather than warnings.
6. Failing baseline check: baseline construction now fails with an actionable message when any declared check guardrail is unsatisfied (matched to targets.json via source_text; a command transcribed nowhere also fails). "Must keep passing" now presumes passing at the baseline.
7. Fallback conversion: a template-fallback conversion refuses --submit outright and points at the YAML to hand-edit, mirroring the faithfulness gate.
8. --force-fresh: the previous workspace is moved aside to a timestamped stale directory before adoption, so re-adoption and a fresh baseline actually happen (and the flag is applied only in the preparation phase under two-phase dispatch).

On the non-blocking statement-invariant concern: you were right that the language overpromised. The schema and the prompt banner now say explicitly that statement invariants are binding instructions but not mechanically verified, and recommend protected_path/check whenever the constraint can be stated as a path or command.

Frankbest18 (changes requested):
Thanks for the detailed update. There are five remaining issues that might need to be resolved:

### P1 — AutoResearch iterations currently crash

The isolated scorer is created and passed into `continue_from_current_best()`, but it is not passed onward to `run_autoresearch_loop()`. That function then refers to `scorer_override` even though it never received or defined it. This produces a `NameError` when an iteration starts.

### P1 — The sealed store remains accessible inside the research container

The sealed data is stored under:

```text
/workspaces/.sealed_store/<idea-id>/
```

However, Docker mounts the entire `/workspaces` directory into the research container. The agent may therefore access the sealed store as a sibling of its project workspace, particularly under the default full-permissions mode.

### P1 — In-repository sealed data remains in Git history

When a sealed dataset is already inside the source repository, NeuriCo removes the original Git history but then creates a new commit before removing the sealed file. Then the file disappears from the visible workspace afterward, but its contents remain in the new commit and can be recovered with a Git command.

The initial GitHub push also occurs before resource staging, so this commit may already have been uploaded.

### P1 — `--force-fresh` can reuse and expose sealed data

`--force-fresh` moves the old project workspace but leaves its sibling sealed store in place.

During the fresh run, staging sees the old stored dataset and treats it as already staged. It can therefore keep the previous dataset instead of using the updated one. For an in-repository dataset, the newly adopted copy may also remain visible because the removal step is skipped.

The result can be stale data in the sealed store and updated held-out data visible in the research workspace.

### P2 — Protected paths do not detect permission-only changes

The protected-path guard hashes file contents but does not include file permissions. Changing only a protected script’s executable bit produces the same fingerprint, so the guard reports no violation even though Git can record that change in the accepted checkpoint.

AndrewRqy:
Thanks a lot for the feedback! All issues should be addressed below:

- Iterations crash (NameError). Fixed. run_autoresearch_loop referenced scorer_override but never declared it as a parameter, and continue_from_current_best did not forward it. The parameter is now declared and passed through, so the isolated continuation scorer is used on every iteration.

- Sealed store readable in the research container. Fixed, and the mechanism is worth spelling out because relocating the store does not solve it. The isolated scorer runs in the same research container, so the store has to be mounted there for scoring to work, which is exactly why the agent can see it. Rather than hide the store, its contents are now ciphertext. Sealed data is Fernet-encrypted at rest, and the scorer decrypts transiently into the throwaway scoring worktree, so the live workspace never holds plaintext held-out data. The key is a launch secret (NEURICO_SEAL_KEY) that the runner reads once and pops out of os.environ before any agent is spawned, so no agent subprocess (os.environ.copy()) inherits it and it is never written under the workspace. When it is unset a random per-user key is generated into the run .env, so nothing is hardcoded or shared and the user can still override. This closes the reward-hacking exposure. It does not defend against an agent scraping the runner's memory for the key, which I am treating as out of scope for this threat model. The escalation, if that ever changes, is to run scoring in its own container.

- In-repo sealed data in git history. Fixed. Adoption committed and pushed before staging removed the in-repo copy, so the anchor commit and the initial GitHub push both carried the bytes. Adoption now extracts the sealed data to the encrypted store and removes it from the tree before the anchor commit and before any push, so neither ever contains it. This retires the "scrubbing history is out of scope" note the old code carried.

- force-fresh reuses and exposes sealed data. Fixed. The store is a sibling of the workspace, so moving the workspace aside left it in place and staging treated the old dataset as already staged. force-fresh now moves the store aside on the same stale suffix, so a fresh run re-stages, and combined with the history fix the freshly adopted in-repo copy is removed too.

- Permission-only changes to protected paths. Fixed. The fingerprint hashed file contents only, so a chmod-only change produced the same digest. It now folds in the file mode, so an executable-bit flip is detected.

One note on the deeper fix for #2. Encryption is the threat-model-appropriate answer, but the airtight version belongs to a larger change: scoring dispatched to its own container so the store is never mounted alongside the agent. Happy to open an issue for that if you agree it is worth tracking. The encryption stays useful either way, and it also covers the local backend, where there is no container isolation at all.

Tested with the encryption round-trip, the store proven to carry no plaintext, the scorer proven to see plaintext only inside its context, adoption asserted to leave no sealed bytes in git log -p or the working tree, force-fresh asserted to clear the sealed store, the chmod-only fingerprint change, and key provisioning into .env without landing in os.environ. 48 tests pass, including the existing suite.

Frankbest18 (changes requested):
Thanks for the update, Andrew. Here are some remaining concerns related to encryption:

### P1 — Automatically generated keys do not persist across Docker phases

Sealed runs use separate preparation and research containers. When no key is configured, the preparation container generates one and encrypts the dataset, but that key disappears when the container exits. The research container then generates a different key and cannot decrypt the stored data.

### P1 — A failed encryption can leave exposed plaintext

The dataset is copied into the sealed store before being encrypted. If encryption fails or is interrupted, plaintext may remain there. On retry, the existing destination is treated as already staged without confirming that encryption completed, potentially making the held-out data readable inside the research container.

### P2 — Large sealed files may exhaust memory

Encryption and decryption load each complete file into memory. A large dataset may therefore require several times its size in RAM and fail during staging or scoring.

AndrewRqy:
Thanks for catching these. All three are fixed and pushed, and i run an additional adversarial review pass on the fix set afterward, which caught a few more issues that are also fixed in the same push.

- Keys not persisting across Docker phases. Fixed at the host. docker/run.sh now provisions NEURICO_SEAL_KEY into the host .env before any container launches, so every phase container receives the same key through the existing --env-file. Inside a container, ensure_seal_key refuses to generate (a key born in a phase container dies with it) and points at the host-side fix. Detection covers docker and podman, and deployments that bind-mount the install, where the container's .env genuinely persists, can opt back in with NEURICO_ALLOW_CONTAINER_KEYGEN=1.
- Failed encryption leaving readable plaintext. Fixed structurally rather than patched: staging now encrypts straight from the source into the store, so the store never holds a plaintext byte at any point. Each file lands via temp write plus atomic rename, and a directory dataset is built in a .staging sibling and renamed as a whole, so a crash at any point leaves either complete ciphertext or nothing. The idempotent re-stage path additionally verifies existing entries are ciphertext (prefix check) and re-stages anything poisoned by the pre-fix code, but only when the source is still reachable, so a poisoned copy that is the last remaining copy is preserved and reported instead of deleted.
- Memory exhaustion on large files. Fixed with a chunked format: a magic header, then length-prefixed Fernet tokens of 4 MB chunks. Peak memory is a couple of chunks regardless of dataset size. Each chunk carries its index and the stream ends with an explicit end marker, so truncated ciphertext fails decryption even when cut cleanly at a frame boundary, as does reordered or trailing data. Files sealed by the earlier whole-file code still decrypt.

I also find the following issues and fixed them on this commit:

- The new encrypt-from-source path initially dropped symlinked files and directories that the old copytree dereferenced, which would have silently truncated held-out datasets with internal symlinks (HuggingFace-style layout followlinks, guards against link cycles, and treats a broken linkas a hard error.
- The ciphertext verification treated an empty directory as valid; and re-stages.
- Key provisioning goes through config_set_env, so an empty NEURICO_SEAL_KEY= placeholder line gets replaced instead of silently blocking generation forever.
- The legacy decrypt path now has the same temp-plus-rename atomic

Frankbest18 (changes requested):
Thanks for the update, Andrew. There are three remaining concerns can might need to be resolved:

### P1 — The encrypted dataset can still be altered by the agent

The sealed store remains writable inside the research container. Although the agent cannot read encrypted contents, it can delete, rename, or move complete ciphertext files. NeuriCo does not verify the expected set of files before scoring, so evaluation may silently use an incomplete or altered dataset. I reproduced deleting one encrypted example and scoring continued with the remaining files.

### P2 — Failed materialization can leave plaintext behind

Files are decrypted before the cleanup `try/finally` begins. If one file decrypts successfully and a later file fails, the earlier plaintext remains under `data/.test`. Isolated continuation scoring has additional cleanup, but the general live scoring path can retain these files after failure.

### P2 — Valid symlink aliases can be omitted

Cycle detection globally skips any directory whose resolved target was previously visited. Consequently, two legitimate symlinks pointing to the same directory are treated like a cycle, and the second path silently disappears from the staged dataset. I reproduced this with two aliases where only the first was preserved.

AndrewRqy:
After some rethinking of the structure for continue-research, I think encryption of the held-out resources may not be the best way to handle this. Instead, continue-research should be treated as a variant of the normal AutoResearch loop, and whenever new evaluation materials are presented, it is necessary to rerun the rule maker so the workspace's scoring protocol is updated properly. So I made the following changes:

1. Removed the held-out encryption. Sealed datasets now go through the same scoring seal the normal loop already uses: they stage as plaintext into data/.test (already listed in SEALED_PATHS), get relocated to .scoring_sealed while any agent runs, and are copied into the frozen scorer worktree only for scoring. No key, no separate store. This drops the whole encryption layer (about 600 lines) and makes the earlier discussion about the key lifecycle and chunked format moot.
2. Rule maker reruns when the scoring protocol is stale. The bootstrap rule maker now regenerates the protocol whenever the workspace has no valid protocol or the declared evaluation materials change, not only on first adoption. It extends the prior protocol (read from a trusted sealed source) and re-baselines the current best under the new protocol, so the workspace's scoring stays consistent with whatever materials are declared.
3. Baseline scoring is isolated when sealed data is present. It goes through run_isolated_scorer, like the iteration path, so held-out data does not sit in the live workspace during baseline scoring.
4. Hardening on the above. The re-baseline restores the workspace to the validated current_best_sha before checkpointing, so an agent-mutated tree cannot become the new baseline. Regeneration reads the idea and prior protocol from trusted sources only, with a no-weakening target floor on retained metrics.

I also rebased onto current main and cleaned the history to three commits. One thing I left alone on purpose: the scoring seal is a relocation to a sibling in the same mount, so a determined agent can still reach it, and setup agents see data/.test before the stage-3 seal. Those are properties of the shared seal already on main (the pipeline_orchestrator comment notes it as deferred to v1.0), not introduced here, so I think they belong in a joint isolation follow-up rather than this PR.

Frankbest18 (changes requested):
Thanks for the update. Removing the encryption layer clears up the earlier issues. I still see four potential problems in the new flow:

### P1 — `sealed: true` does not fully hide held-out data

The dataset is first stored as plaintext in `data/.test`. The manifest-trimmer and rule-maker agents run before that directory is sealed, so they can read it despite being instructed not to.

Later, the data is moved to `.scoring_sealed`, but that directory is still inside the same writable `/workspaces` mount. An agent can reach it by navigating outside its workspace. It can also modify both the dataset and its integrity manifest.

### P1 — Changes to evaluation data may not take effect

When a new sealed dataset is added, NeuriCo stages it and detects the change. Re-baselining then restores the previous checkpoint and deletes `data/.test`, including the newly staged dataset. It is not staged again before scoring.

There is a separate problem when replacing a dataset but keeping the same name. NeuriCo sees the existing staged destination and keeps the old copy. Because the change fingerprint only considers the dataset name, changing its source or contents also does not trigger regeneration.

As a result, scoring may use missing or outdated evaluation data.

### P1 — Regeneration may lose the previous scoring protocol

The previous protocol is read only from `.scoring_sealed`. In a normal completed run, those files have already been moved back into the workspace and the sealed directory has been removed.

During regeneration, NeuriCo removes the workspace scoring files and then looks for the previous protocol in the missing sealed directory. The rule maker can therefore receive no prior protocol to extend.

The target-floor check also ignores an old metric if the regenerated protocol drops it completely. Existing requirements can disappear without being reported as weakened.

### P1 — A failed evaluator can be accepted as the baseline

`run_scorer()` correctly reports failure when `eval.py` exits with an error, even if it wrote a partial `results.json`. The isolated baseline wrapper then treats the run as successful if a results object and scored checkpoint exist.

A baseline is allowed to miss its targets, but the evaluator itself still needs to finish successfully. Otherwise, incomplete results written before a crash can become the baseline for later comparisons.

AndrewRqy:
Thanks for the review. All four are confirmed and fixed on the branch:

P1, sealed data not fully hidden. The plaintext window is closed: data/.test is now sealed out of the workspace before Pass 1 of the manifest stage and stays sealed through the manifest trimmer, rule maker, and eval verifier, so no setup agent can read held-out data while it authors or checks the harness. Manifest entries for the sealed datasets are synthesized from the sealed location by trusted code, so agents still see the datasets exist. The data is restored just before the scorer, and a crash inside the sealed window strands a copy that the next run adopts and restores (staging recognizes the parked location too). Your second and third points, that the sealed sibling is still ..-reachable on the same mount and the integrity manifest is co-located with the data, are correct and are exactly the off-mount isolation work: as discussed, that is scoped out of this PR and tracked as a follow-up, since it needs an agent-isolation boundary rather than a relocation.

P1, eval-data changes may not take effect. Fixed in three parts. The rebaseline restore no longer wipes the just-staged dataset: freshly staged data/.test is preserved across restore_checkpoint(remove_hidden_scoring=True) and is present for baseline scoring. Same-name replacement now re-stages: staging records a content hash per sealed dataset at stage time, and when the declared source is readable with a different hash it replaces the staged copy, including any seal-relocated stale copy. And the materials fingerprint now includes each dataset's content hash (recorded hash as fallback once the source is gone), so changing contents or source under the same name triggers regeneration. One note: the fingerprint format change means the first resume of a pre-existing workspace after this lands will re-fire one rebaseline.

P1, regeneration loses the previous protocol. The prior is no longer tied to the transient seal. Every time a protocol is validated and its baseline accepted, the runner persists scoring/{eval.py,targets.json,interface.md} to a durable .protocol_store sibling, and read_prior_scoring_protocol reads the sealed copy first and the store second, never the worker-writable workspace copy. So a completed-run lifecycle now hands the rule maker a real prior to extend. On the floor check: a retained metric that is dropped entirely from the regenerated targets is now reported as a violation instead of silently skipped, so requirements cannot disappear unflagged.

P1, failed evaluator accepted as baseline. Fixed. The isolated baseline wrapper no longer ORs away run_scorer's result; stage success is exactly run_scorer's success, which encodes a clean eval.py exit with results and no error. A crashed evaluator that left a valid-JSON partial results.json can no longer anchor the baseline. A baseline may still miss its targets, as intended.

Frankbest18 (changes requested):
Thanks for the update. There are four remaining concerns that might need to be resolved:

### P1 — Setup agents can run while sealed source paths are mounted

In Docker, the preparation container receives the original sealed host paths and is supposed to exit before any agents run. On a resumed workspace, however, NeuriCo handles protocol regeneration before checking `--prepare-workspace`.

If evaluation materials changed, the manifest trimmer, rule maker, and verifier can therefore run inside the preparation container while the original held-out source remains readable. Moving `data/.test` does not hide that original mount.

### P1 — The durable protocol store is still agent-writable

The new `.protocol_store` is stored under the same writable `/workspaces` mount available to agents. An agent can modify the stored `eval.py` or `targets.json`, and those changes survive normal workspace rollback because the store is outside the project repository. A later regeneration then reads that modified copy as the trusted prior protocol.

### P2 — Content-hash restaging is not fully reliable

For a single-file dataset, the source hash is recorded before copying finishes. If the copy is interrupted and leaves a partial destination, a retry sees that the destination exists and that the source still matches the cached source hash. The partial destination itself is not checked, so it can be kept as complete.

There is also a mismatch for directory symlinks: staging follows them and copies their contents, but the hashing walk does not include files behind directory symlinks. Changes to those files may therefore leave the old staged dataset and old scoring protocol in use.

### P2 — Dropped-metric validation still has a fail-open case

The new check correctly detects a missing old metric when the regenerated protocol contains valid numeric targets. However, the structural validator does not require every property target to be numeric.

If all regenerated targets are malformed or nonnumeric, the numeric target map is empty and the target-floor check returns no violations before comparing against the prior metrics. This treats “unable to validate the new targets” as “no regression found.”

Issue #51, opened 2026-03-03
Context management and working memory


laoliu5280:
Agents lose track during long runs. They forget earlier instructions, drift off task, or even lose their working directory.

### Tasks

- [ ] Maintain a `STATE.md` file in the workspace that the agent updates at each phase (current phase, what is done, key findings, next steps)
- [ ] Add periodic working directory checks (agents losing their directory is a known failure mode)
- [ ] Before each phase, generate a summary of prior phases to keep the context focused
- [ ] Validate that each phase produced its expected outputs before starting the next one
- [ ] Limit the agent to the top-K most promising directions instead of exploring everything

### Research questions

- What information matters most across phases?
- How to detect context drift during execution?
- What is the right tradeoff between context breadth and depth?

PR #144, opened 2026-07-24
Add durable phase handoffs and output validation

Lanoxia:
## Summary

Addresses #51 by giving each pipeline phase a durable, inspectable handoff contract. Agents can resume from `STATE.md` without losing decisions, while the orchestrator verifies workspace identity and expected artifacts at phase boundaries.

## What changed

- Add atomic `STATE.md` generation with current phase, resume summary, next steps, workspace checks, output validation, and preserved agent-owned notes.
- Add workspace and output validation, including rejection of expected outputs that resolve outside the workspace.
- Gate resource finding, experiments, rule making, scoring, and bootstrap stages on valid workspace/output state.
- Add an agent state contract and configurable top-K research direction budget (`idea.max_directions`, default `3`, range `1..10`).
- Document the workflow and add schema/validation coverage.

## Verification

- `python -m pytest tests -q` -> `76 passed`
- `python -m py_compile src/core/phase_state.py src/core/pipeline_orchestrator.py src/core/idea_manager.py src/templates/prompt_generator.py`
- `git diff --check`

The repository tests are clean. The unrelated recursive `services/paper-finder` suite was not included because this environment lacks its separate `ai2i` and `langchain` dependencies.


AndrewRqy (changes requested):
Thanks a lot for the PR :) This looks very clean and well structured. There are a few points i want to bring to attention:

1. Consider dropping the redundant (and stale-prone) resume summary. STATE.md currently renders the prior-phase roll-call twice: "Previous phases" (recomputed live at render) and "Current phase context → Resume summary" (frozen into the stage record at start_stage). They carry the same information, and the frozen copy goes stale, if a prior phase's status changes after a later phase starts (e.g. a re-run), the two disagree and STATE.md contradicts itself, showing the same phase as both succeeded and failed. Simplest fix is to remove the resume_summary field/section and keep the live "Previous phases" list as the single source.

2. Have resource_finder record its own notes in STATE.md. Right now the state contract is only injected into the experiment_runner's prompt, so it's the only agent that writes into the notes block. That means resource_finder, the first substantive phase, contributes nothing to the handoff, its findings live only in literature_review.md/resources.md, and the one agent that does write notes (experiment_runner) is the last phase, so intra-run the notes never actually reach a later phase. Could you distribute the contract to resource_finder as well, so it leaves its own notes (key findings, the directions it kept/pruned, recommended next steps) for experiment_runner to read? To avoid the two agents overwriting one shared block, it's probably cleanest to give STATE.md a per-phase notes section (e.g. a labeled block per stage) and tell each agent to own only its own section.

Do you mind keeping only one phase tracker to reduce redundancy and maybe enable resource finder to add notes to the state.md file as well? Everything else looks great.

AndrewRqy (approved):

PR #146, opened 2026-07-27
feat: add manager-mediated HITL AutoResearch workflow with durable workspace state

Frankbest18:
## Summary

This PR consolidates the earlier exploratory HITL work from #135 and #141 into a separate and end-to-end HITL AutoResearch workflow.

The main change is the ownership model:

- workers perform research and interact with HITL through runtime commands;
- the manager reviews plans, raised questions, completed work, proposals, scores, and frontier choices;
- the human provides approval or feedback when the manager escalates;
- runtime owns official HITL state, phase transitions, scoring handoffs, recovery boundaries, and frontier persistence.

Ordinary `--autoresearch` and `--continue-autoresearch` retain their existing behavior. The HITL workflow is entered only through the new HITL-specific modes.

This is a large integration PR. After several rounds of succesful local end-to-end provider-backed multi-iteration smoke testing, the implementation is ready for architectural review .

## Feedback needed

The areas where feedback would be most useful are:

1. **Responsibility boundaries:** Is the current split between worker, manager, human, and runtime clear and appropriate?

2. **Research-state boundaries:** Are the roles of the idea log, frontier, whiteboard, manager conversation, and `ResearchState` sufficiently distinct?

3. **Frontend design:** Are current frontend design reliable, stable and easy to use now? Most tests were based around the web page, so further feedback on the client side is needed.

### Dedicated HITL AutoResearch modes

New entry modes:

```bash
--hitl-autoresearch {web,cli}
--hitl-continue-autoresearch {web,cli}
```

These modes use a separate HITL controller. They do not change the control flow of ordinary AutoResearch.

The HITL path covers:

- `resource_finder`
- `rule_maker`
- AutoResearch proposal generation
- experiment planning and execution
- semantic phase review
- scoring
- candidate accept/reject decisions
- frontier pruning and selection
- restart and rollback recovery

### Runtime-mediated worker commands

Workers no longer communicate HITL completion or escalation through ad hoc files or process exit alone. The shared command contract is:

```text
material autonomous evidence or decision
    -> hitl-report-idea

cannot continue without manager or human resolution
    -> hitl-raise-idea

phase work is ready for review
    -> hitl-finish-phase

AutoResearch proposal is complete
    -> hitl-submit-proposal
```

Runtime validates each command and fills runtime-owned fields such as identity, stage, actor, provenance, and decision level.

Workers are instructed not to edit `.neurico/hitl/` directly. They receive only the commands that are legal for their current phase, and runtime validates the command surface again when a command is invoked.

### Living plans and explicit phase completion

Stage work is divided into plan, execution, and review/revision phases. The stage worker owns its living plan and remains responsible for revising and executing it after manager or human feedback. Replacement workers receive the same complete stage context as the original worker.

A provider process exiting is treated only as a liveness event. A phase completes only after:

1. the worker calls `hitl-finish-phase`;
2. runtime validates the workspace and artifact boundary;
3. the manager completes the required semantic review;
4. any required scoring or frontier transition is durably finalized.

External workers also run in isolated process groups. If the provider exits while leaving background child processes that can continue modifying the workspace, runtime terminates those children and does not treat the invocation as successful.

### Long-running HITL manager

The HITL manager is a persistent ReAct agent with one chronological conversation and one runtime-controlled tool loop.

The manager has general read-only tools for:

- workspace listing, search, and file inspection;
- viewing finalized HITL ideas;
- recalling earlier manager conversation;
- inspecting the active frontier;
- updating higher-level research synthesis.

Mutating tools are exposed only at the runtime boundary where they are legal. For example:

- `finalize_worker_request` is available for a held worker request;
- `approve_for_scoring` is available only at a scoring approval boundary;
- `finalize_frontier_decision` is available only after runtime-owned score evidence exists;
- `select_frontier` and `prune_frontier` are available only at their dedicated frontier boundaries.

Runtime provides the tool surface through a private local MCP bridge and an allowlist. The current manager path supports Codex and Claude CLI backends.

### Manager context and long-term recall

Manager memory has two layers:

- an active JSONL conversation context with bounded, recursive compaction;
- a complete SQLite transcript archive with FTS-based retrieval.

Older tool outputs can first be replaced with bounded placeholders. When additional compaction is needed, older conversation is summarized while recent turns remain available verbatim. The full raw history is retained separately and can be retrieved through `recall_manager_conversation`.

### Human conversation and resolution replies

When human input is needed, the manager asks a focused question, receives the human response, and translates it into the structured feedback needed by the waiting worker.

A human resolution reply is separate from ordinary chat and is tied to the active runtime request key. This prevents unrelated conversation or a stale browser request from resolving the wrong worker command.

Ordinary conversation with the manager remains available while a worker request is pending.

### Logical research record

The official HITL audit trail is:

```text
.neurico/hitl/idea/idea.jsonl
```

It supports three record types:

- `evidence`
- `decision`
- `proposal`

Workers provide the substantive content, while runtime owns identifiers, timestamps, stage attribution, actor, level, raised status, and AutoResearch provenance.

Evidence may begin a new lineage. Decisions and proposals must cite already-finalized premises. Because records may cite only earlier records, the premise structure forms a DAG without introducing a separate graph database.

An idempotent synchronization layer projects source-linked information from the idea log, frontier, and whiteboard into `ResearchState`. `ResearchState` remains a manager-facing synthesis rather than a second authoritative HITL store.

### Frontier-based AutoResearch

HITL AutoResearch maintains:

- one selected frontier node;
- a portfolio of active frontier nodes;
- accepted and rejected attempt histories;
- the objective score and manager rationale for each finalized attempt.

Proposals are classified as:

```text
exploitation
exploration
```

Accepted exploitation replaces its parent as the current node for that research direction and inherits the direction's earlier attempt history; accepted exploration keeps both the parent and child active, creating a separate direction.

Rejected candidates remain in the parent node's attempt history as valid research evidence. Pruning removes a node from the active portfolio but does not delete its audit history.

The proposer can inspect the currently selected direction. The manager can inspect the full active portfolio and is responsible for pruning and selecting the direction used by the next proposal.

### Proposal admission

AutoResearch proposals are submitted directly to runtime through `hitl-submit-proposal`.

The admission path is:

```text
proposal submission
    -> manager legality review
    -> human approval or feedback
    -> materialized experiment plan
```

The manager checks for evaluator manipulation, direct score manufacture, protected-boundary violations, and other reward-hacking behavior before the proposal is shown to the human.

A rejected proposal is replaced with a new proposal rather than editing an already-submitted record in place.

### Runtime-owned scoring

The rule maker establishes evaluator authority once. Runtime then seals evaluator internals outside the public worker workspace, including the evaluation script, targets, selected rule-maker artifacts, and hidden evaluation data. The sealed payload receives a manifest with hashes and required entries.

For candidate scoring, runtime:

1. verifies that the public workspace still matches the state reviewed by the manager;
2. checkpoints the reviewed candidate;
3. creates a detached private Git worktree at that checkpoint;
4. copies the sealed evaluator into the private scorer workspace;
5. runs the scorer there;
6. commits the score result against the immutable candidate source;
7. retains the scored checkpoint through a temporary private Git ref;
8. publishes `scoring/results.json` to the public workspace as a review copy.

The public result file is not the scoring authority. The manager receives the runtime-owned scorer result and checkpoint provenance.

If a worker recreates or changes sealed evaluator files in the public workspace, runtime rejects the attempt rather than resealing the worker-created replacement.

### Durable recovery

Private Git refs store rollback boundaries for HITL state separately from public AutoResearch node checkpoints.

The durable state includes:

```text
.neurico/hitl/
.neurico/research_state.json
```

Live locks, temporary files, generated command wrappers, MCP credentials, and SQLite sidecars are excluded from snapshots.

`.neurico/hitl/runtime.json` records transitions that must survive process restart, including:

- worker continuation state;
- one pending worker command;
- plan approvals;
- scoring handoffs;
- frontier pruning or selection actions;
- rejected-attempt whiteboard cleanup;
- frontier-decision publication.

Frontier publication is persisted as an idempotent sequence:

```text
prepared
    -> idea_logged
    -> frontier_finalized
    -> mirrored
    -> completed
```

A restart resumes the incomplete transition instead of asking the manager to repeat the same decision.

The workflow distinguishes:

- **failed or interrupted attempts**, which are rolled back and removed from canonical research history;
- **scored but rejected attempts**, whose score, rationale, and useful cross-attempt learning are retained.

Failed attempts leave only a small `runtime_incident.json` diagnostic record. Partial failed work is not presented as a valid experiment.

### Dedicated web and CLI interfaces

The HITL web interface is separate from the ordinary interactive-manager page, it reads complete snapshots derived from canonical workspace artifacts. Server-sent events are used only as refresh notifications; the browser does not reconstruct research state from event fragments.

The interface displays:

- manager conversation;
- queued human messages;
- the active resolution request;
- finalized ideas;
- selected and active frontier nodes;
- accepted and rejected attempts;
- cross-attempt whiteboard lessons;
- `ResearchState`;
- manager context usage;
- chronological HITL activity.

The same interface can launch a fresh HITL AutoResearch run or continue an existing frontier.

## Example usage

Fresh HITL AutoResearch in the web interface:

```bash
./neurico run <idea_id> \
  --provider codex \
  --hitl-autoresearch web \
  --autoresearch-iterations 3
```

Continue an existing HITL frontier:

```bash
./neurico run <idea_id> \
  --provider codex \
  --hitl-continue-autoresearch web \
  --autoresearch-iterations 3
```

Use the terminal interface:

```bash
./neurico run <idea_id> \
  --provider claude \
  --hitl-autoresearch cli \
  --autoresearch-iterations 3
```

Open the HITL workspace page independently:

```bash
./neurico hitl-web <idea_id>
```

### Runtime and AutoResearch

- `src/core/hitl.py`
- `src/core/hitl_autoresearch.py`
- `src/core/hitl_runtime_state.py`
- `src/core/hitl_frontier.py`
- `src/core/hitl_git_state.py`
- `src/core/hitl_scoring_workspace.py`
- `src/core/hitl_workspace_guard.py`

### Manager and memory

- `src/core/hitl_manager_react.py`
- `src/core/hitl_manager_context.py`
- `src/core/hitl_manager_history.py`
- `src/core/hitl_manager_host.py`
- `src/core/hitl_manager_mcp.py`
- `src/core/hitl_world_model.py`

### Human interfaces

- `src/cli/hitl_web.py`
- `src/interactive/hitl_web_server.py`
- `src/core/hitl_workspace_view.py`
- `src/interactive/static/hitl/`

### Worker contracts

- `templates/hitl/`
- `src/agents/resource_finder.py`
- `src/agents/rule_maker.py`
- `src/agents/comment_handler.py`
- `src/agents/autoresearch_proposer.py`

## Validation

The branch adds focused tests for:

- runtime worker commands and stage transitions;
- manager tool authorization and ReAct behavior;
- manager conversation compaction and SQLite recall;
- proposal admission;
- frontier acceptance, rejection, pruning, and selection;
- restart and rollback behavior;
- private Git-backed HITL state;
- evaluator sealing and isolated scoring;
- workspace write guards and artifact inspection;
- web input, snapshots, and security boundaries;
- `ResearchState` synchronization;
- whiteboard persistence and rejected-attempt cleanup.

## Compatibility and current limits

- Ordinary AutoResearch remains unchanged.
- HITL does not automatically transition into or out of ordinary AutoResearch.
- The private Git recovery refs are local to one repository and are not transferred by ordinary push or clone.
- Scoring isolation is a correctness and evaluator-exposure boundary, not an OS-level sandbox.
- There is no compute-backend-specific HITL behavior in this PR.
- Required-artifact validation is enforced at phase boundaries, but individual artifacts do not yet have `must_create`, `must_change`, or `may_reuse` freshness policies.


AndrewRqy:
Thanks a lot for the PR, here are some of the suggestions I have for the runtime core part (hitl.py, hitl_autoresearch.py, hitl_runtime_state.py, hitl_git_state.py, hitl_frontier.py, the worker command handlers). 

1. The frontier decision resume path is dead code due to an attempt id format mismatch.

The transition record stores the bare attempt id: attempt_id = attempt_dir.name (hitl_autoresearch.py:1415), passed through _run_candidate_experiment_hitl (1444) into _prepare_frontier_decision (1789). But both resume-side consumers compare against the full <parent>/attempt_N form:

- recovery classification compares transition.attempt_id == marker (line 336), so frontier_decision_transition recovery is never selected and _resume_frontier_decision_transition is unreachable;
- _complete_scored_hitl_attempt compares against self._attempt_id(attempt_dir) (line 1170), so the completed advance never fires and every transition ends life at mirrored.

Consequence: a crash inside the publication window (prepared through mirrored) is misclassified on restart. Depending on the pending command state it either relaunches the worker or takes the full rollback path, which restores .neurico/hitl and erases the already-finalized manager decision, then re-asks the manager on retry. The public mirror under history_root/nodes is not reverted, so the audit mirror diverges. This contradicts the stated behavior that a restart resumes the incomplete transition instead of repeating the decision. The fix is small: use one format at the write site (1415) or the three comparison sites, and add a resume test that crashes between each pair of transition states.

2. Resuming a mid-scoring crash spuriously rejects the reviewed candidate.

The reviewed-state fingerprint is captured at worker finish, before scoring. _clear_stale_results_json (1726) then deletes the tracked scoring/results.json that existed when the fingerprint was recorded. On resume with isolated_scoring.status == "prepared", the fingerprint re-check (1706-1718) sees a changed workspace and discards the candidate as tampered. Either exclude scoring/results.json from the fingerprint or re-capture the fingerprint after the stale file is cleared.

3. Scoring ref lifecycle can leak and then block future attempts.

- Module-level recovery (376-394) restores the .neurico/hitl snapshot without retiring the temp scoring ref first. The only pointer to refs/neurico/hitl/scoring/<request_key> lives in runtime.json, so restoring the snapshot orphans the ref permanently.
- Failed attempts remove their attempt dir (1116, 1495, 1568) and next_attempt_dir reuses max+1 numbering, so a leaked rollback ref later collides: begin_autoresearch_hitl_attempt hard-raises on an existing ref (hitl_git_state.py:73-76), which blocks every subsequent attempt for that parent until manual ref deletion.
- _retire_temporary_scoring_ref(strict=True) (1307) raises on a missing ref, so a replay of _commit_frontier_decision after a crash that happened post-retire cannot complete. Retire should be idempotent.

4. _run_iteration_until_scored retries without a cap.

Lines 770-783 loop until a scored result. A deterministic pre-scoring failure, for example a seal error at 1429, loops forever and archives a new incident each pass. The non-HITL loop caps this with MAX_INVALID_ATTEMPTS_PER_VALID_ITERATION = 3 (autoresearch.py:81). The HITL loop should have an equivalent bound.


Duplication

There are two patterns worth fixing before merge, since both will cause drift.

Re-implemented utilities. The same low-level concerns are hand-rolled repeatedly across the new files:

- Atomic write (tmp + fsync + os.replace + dir fsync): 9 copies (hitl_runtime_state.py:77-106, hitl_frontier.py:67-109 twice, hitl.py:478-492, 1571-1578, 3746-3755, hitl_manager_context.py:240-254, hitl_manager_inbox.py:57-64, hitl_autoresearch.py:264-276, hitl_scoring_workspace.py:43-60), with three different tmp-naming schemes.
- UTC timestamp helper: 5 implementations in 5 formats (hitl.py:112, hitl_runtime_state.py:23, hitl_manager_context.py:30, hitl_manager_inbox.py:20, plus the naive-local calls in hitl_autoresearch.py). Two are timezone-naive, which is a sorting bug, not just style.
- _sha256_file: 4 new copies (hitl.py:3841, hitl_scoring_workspace.py:27, hitl_workspace_guard.py:174, hitl_workspace_inspection.py:376) when scoring_seal.py:27 already has it, and hitl_scoring_workspace.py already imports from scoring_seal.
- Git subprocess plumbing: 13+ raw call sites across hitl_git_state.py, hitl_frontier.py, hitl_scoring_workspace.py, hitl_autoresearch.py with three different error policies (hitl_git_state.py:414-425 discards stderr, hitl_frontier.py:217-227 surfaces it). hitl_autoresearch.py contains two separate git update-ref -d wrappers (213 and 1307).
- .neurico/hitl path literals hard-coded in 10 files even though hitl.py defines hitl_state_dir (159).
- Two readers of idea.jsonl with different error handling (hitl.py:3606 vs hitl_world_model.py:100).

Suggested fix: add two small modules. hitl_paths.py holding the path constants and hitl_state_dir, so modules stop hard-coding literals to avoid importing the 3.8k-line hitl.py. hitl_util.py with now_utc(), atomic_write_json/text(), fsync_directory(), read_jsonl(), and run_git(work_dir, *args) raising errors that carry stderr. Move _sha256_file there or import it from scoring_seal. This is mechanical and removes roughly 20 duplicate blocks.

Forked copies of autoresearch.py code. hitl_autoresearch.py already imports 11 primitives from autoresearch.py, but re-implements others verbatim, docstrings included: _ensure_results_json (2043 vs autoresearch.py:1820), _clear_stale_results_json (2075 vs 1866), _idea_with_comments (2069 vs 1860), _attempt_id (2134 vs 1871), _revert_whiteboard_for (2147 vs 1884), and _call_proposal_generator (1988 vs 1788, HITL adds env plumbing). Same for the marker trio in hitl_whiteboard.py:43-62 vs whiteboard.py:101-131. Any future fix to the non-HITL loop will silently miss the HITL fork.

Suggested fix: hoist these to module-level functions in autoresearch.py (or a shared autoresearch_common.py) parameterized where the variants differ (whiteboard class, extra env), and import them from both controllers. Where a method truly needs controller state, a small mixin shared by both controllers does it.

Also worth a deliberate decision: the pre-existing checkpoint code uses GitPython while all new HITL git code shells out to subprocess. Either is fine, but one feature using both styles with four error-handling conventions makes failures inconsistent to debug.

AndrewRqy:
Following up with the second part of the review, covering the integration surface (the modified pre-existing files: pipeline_orchestrator.py, agent_runner.py, runner.py, autoresearch.py, llm_backend.py, the four agents, and templates).

First the good news: the mode gating is clean. Every HITL branch in run_pipeline, runner.py, and the four agents is guarded, the new kwargs all default to legacy behavior, the CLI surface is additive, and the ordinary resource finder and proposer renders leak no HITL command text. The new process-group and timeout semantics in run_prebuilt_cli_agent also apply to ordinary runs, and we are fine adopting that as a standardization (the old streaming-timeout behavior was effectively no timeout).

The one blocking bug

Ordinary AutoResearch is broken on this branch. _resolve_proposal_text lost its @staticmethod decorator (main has it at autoresearch.py:1844; the branch does not at autoresearch.py:1846), while the call site is still self._resolve_proposal_text(attempt_dir, proposal_result) (autoresearch.py:1629). That passes three positional args to a two-parameter function, so every ordinary --autoresearch / --continue-autoresearch iteration raises TypeError. The exception is swallowed by the surrounding try/except Exception (1623-1637) and recorded as a pre-scoring failure, so each iteration runs the paid proposer CLI, then fails the proposal stage, skips the worker and scorer, and is rejected with reason = "AutoResearch proposal/comment stage failed: ...TypeError...". The loop degenerates into N failed candidates while spending proposer tokens.

The fix is one line (restore the decorator or add self). The larger point is what made it possible: HitlAutoResearchController re-implements AutoResearchController rather than sharing code with it, so a change made with the HITL consumer in mind went out without the ordinary call site being updated, and with the tests excluded from the branch nothing caught it. A regression test that runs one ordinary iteration end to end with stubbed agents would have flagged this immediately.

Small correction to my previous message: I cited MAX_INVALID_ATTEMPTS_PER_VALID_ITERATION as the non-HITL loop's retry cap. It is defined (autoresearch.py:81) but referenced nowhere, so neither loop currently has a cap. The suggestion to bound the HITL retry loop stands; wiring this constant into both loops would cover it.

Duplication

- The controller fork. HitlAutoResearchController duplicates the checkpoint/history/scoring flow of AutoResearchController instead of extending or composing it. This is the root cause of the bug above and will keep producing drift. Worth deciding on a shared base class or shared module-level helpers before merge, since it only gets harder later.
- Orchestrator internals. _run_resource_finder_hitl, _run_rule_maker_hitl, and _run_experiment_runner_hitl are ~200-line structural clones (checkpoint, snapshot, plan/execution branches, approved/failed epilogues) differing only in worker callable, validator, and prompt source (pipeline_orchestrator.py:724-929, 1213-1550, 1579-1775). A parameterized stage driver would remove roughly 400 lines. Inside them, the rollback closure appears three times (pipeline_orchestrator.py:769-799, 1263-1293, 1618-1650) with a fourth near-copy at _recover_experiment_runner_from_runtime_checkpoint (551-609), and the replacement-worker relaunch loop appears four times (pipeline_orchestrator.py:832, 1323, 1679, hitl_autoresearch.py:1900).
- Agent-side copy-paste. The env_extra merge line is identical in all four agents (resource_finder.py:198, rule_maker.py:303, comment_handler.py:364, autoresearch_proposer.py:321); the ~20-line hitl_runtime launch branch is duplicated between resource_finder.py:212-227 and rule_maker.py:312-325; the prompt-suffix append is duplicated between autoresearch_proposer.py:287 and comment_handler.py:339. One thin helper around run_prebuilt_cli_agent covers all of these.
- Two mechanisms for the same worker contract. The idea-reporting contract is injected via a Jinja include in the proposer template (templates/agents/autoresearch_proposer.txt:41-43) but via the Python helpers (HitlRuntime.idea_reporting_prompt_block, hitl.py:978-987) for every other worker. Same template, same three parameters, two composition paths; the proposer should go through the hitl.py helpers like the rest.
- Smaller: two subtly different process-group kill helpers (agent_runner.py:343-354 uses os.getpgid(pid), llm_backend.py:268-288 calls killpg(pid) directly); run_prebuilt_cli_agent duplicates _run_cli_agent's Popen/stream/sanitize loop in the same file (agent_runner.py:220 vs 153); the skill-root literal is re-derived at prompt_generator.py:929 and pipeline_orchestrator.py:607-611 despite the existing _skill_root_for_provider (prompt_generator.py:551); the artifact validator at pipeline_orchestrator.py:756-767 hand-rolls the same checks as verify_required_artifacts in hitl.py.

build_comment_handler_launch (comment_handler.py:317) is exactly the right pattern, one launch builder shared by the ordinary and HITL paths. Extending that approach to the other agents would resolve most of the agent-side items.

One more suggestion: default model values are now spread across hardcoded strings and three differently named config keys (llm_backend.py:534, 587; llm_model / hitl_manager_llm_model / codex_model), and it would be cleaner to gather them into a single module or config section that everything imports; happy to either see it here or open a separate PR for it ourselves.

AndrewRqy:
Continuing the review with the scoring and workspace-integrity part (scoring_seal.py, whiteboard.py, hitl_whiteboard.py, and the new hitl_scoring_workspace.py / hitl_workspace_guard.py / hitl_workspace_inspection.py). 

1. Reseal can lose sealed artifacts. seal_scoring_files builds the new sealed generation from the current workspace, then shutil.rmtrees the old sealed dir wholesale before publishing (scoring_seal.py:248-251). If a prior unseal partially failed and left a file stranded in the sealed dir while it is absent from the workspace, the next seal deletes it for good. eval.py/targets.json are shielded because their absence trips verification first, but rule_maker_log.md and data/.test/ are not, and get silently lost. The old code merged into the existing sealed dir and never destroyed leftovers. Not urgent, but worth publishing the new generation before removing the old one and reconciling rather than replacing wholesale.

2. A seal failure crashes the whole loop instead of failing one attempt. seal_scoring_files now raises when the payload is missing eval.py/targets.json or contains a symlink (verify_sealed_scoring_manifest + _manifest_entry), where before it sealed whatever subset existed. The call is at autoresearch.py:1622, outside the iteration's try/except that starts at 1623, so the exception escapes run_iteration and takes down the entire AutoResearch loop. The tighter validation is reasonable, but a bad seal on one attempt should be caught and retried (recorded as a failed attempt, loop continues), the same way proposal/comment failures already are, rather than being fatal. Moving the seal inside the failure-handling path, or catching the seal error into a failed-attempt result, would do it.

3. Ordinary iterations run git machinery for history nobody reads. write_current_attempt_marker now calls HitlGitStateStore.begin_autoresearch_whiteboard_attempt (a git commit into a private ref) on every ordinary iteration, and Whiteboard.save() version-records on every write, both adding new raise paths (whiteboard.py:116,147-154,256-258). The resulting refs/neurico/autoresearch-whiteboard history has no readers in ordinary mode; ordinary revert still uses the file-based whiteboard_before.json. This should be gated behind the HITL whiteboard flag along with the rest of it, so ordinary runs neither pay the cost nor inherit the new failure modes.

Scoring isolation

4. The seal does not fully contain ordinary agent activity (flagging, uncertain severity). Three routine actions reach inside the scorer: it runs <worktree>/.venv/bin/python, symlinked to the live worker .venv (hitl_scoring_workspace.py:136-138), so a patched package or a .pth/sitecustomize.py executes; eval.py runs with its own scoring/ directory on sys.path[0] and copy-in overwrites only the four sealed files without clearing that directory, so a co-located module can shadow an import; and seal verification covers only the four manifest paths, not the transitive set eval.py imports. None of this requires adversarial intent, it is reachable by normal experiment work, which is why the seal may not give the containment it advertises. That said, we have not observed any reward-hacking behavior here in practice, so I am unsure how much this matters today; flagging it as something to tighten when convenient rather than a blocker. If you want to close it, running the scorer with an explicit clean env= and a fixed interpreter instead of the workspace .venv, and clearing the scoring/ directory beyond the four named files, would cover it.


Duplication

- _sha256_file is copied five times with identical bodies (scoring_seal.py:27, hitl_scoring_workspace.py:27, hitl_workspace_guard.py:174 plus the fingerprint hasher, hitl_workspace_inspection.py:375). The root cause is that scoring_seal's copy is _-private; exporting it (or moving it to a shared util) and importing everywhere fixes all five. hitl_scoring_workspace already imports from scoring_seal.
- The excluded/hidden path-prefix sets are triplicated and already inconsistent: the guard's _EXCLUDED_PUBLIC_PREFIXES includes logs, inspection's _HIDDEN_PATH_PARTS omits it, and the inspection ripgrep globs are a third copy. These should be one shared constant.
- SEALED_PATHS is re-encoded as inspection's _PROTECTED_RELATIVE_PATHS/_PROTECTED_PREFIXES; they will drift if the seal set changes. Import the single source.
- The marker trio in hitl_whiteboard.py:43-62 looks copy-pasted from whiteboard.py:105-134 but has one real behavioral difference: the HITL variant calls begin_hitl_autoresearch_whiteboard_attempt unconditionally with no .git guard, while the ordinary one silently no-ops. The copy-paste appearance hides the divergence; parameterizing into one helper would make the difference explicit.
- unseal_scoring_files has a strict= parameter (scoring_seal.py:273) that no caller passes, and its docstring claims HITL uses it, but HITL never unseals (it uses the isolated worktree). It can be removed.

AndrewRqy:
Bug: a worker gate can be delivered to a manager that has no tools to resolve it, wedging the run

Observed during a live web run at the resource_finder plan-approval gate:
- The runtime delivered a plan-review request whose own contract instructs the manager to call ask_human and then finalize_worker_request.
- But the manager's provisioned tool set contained only the read-only/synthesis tools; ask_human and finalize_worker_request were absent. So the gate could not be resolved by the manager, and the human could not resolve it either (ordinary chat correctly cannot close a gate, and there was no formal path). The worker stayed paused indefinitely.
- The page gave no signal: the snapshot reported no pending request, so there was nothing on screen to act on while the run was silently stuck.

Root cause, verified from the live runtime.json and hitl_manager_react.py:
- The manager backend became unavailable, its bounded retry budget was exhausted, and the runtime cancelled the pending request: pending_worker_command.status is cancelled, with reason "The HITL manager backend remained unavailable after its bounded retry budget ... Runtime is rolling back this AutoResearch attempt," cancelled_at two seconds after created_at.
- _available_tool_names (hitl_manager_react.py:392-393) surfaces ask_human/finalize_worker_request only when the pending command status is exactly pending. Once it flipped to cancelled, those tools were gone for good, which is exactly the read-only-only tool set we observed.
- The "rollback" did not actually abandon the attempt. worker_continuation.status was replacement_running with replacement_count: 5, and a worker subprocess was still running, so the run kept relaunching resource_finder workers, each a full agent run spending tokens, none of which could ever be resolved.

Why this is high severity: the trigger is any manager-backend unavailability, a transient rate limit, a timeout, or a misconfigured/missing backend, all of which are ordinary. In our run the trigger happened to be that the default hitl_manager_provider is codex (hitl_manager_react.py:266) and codex was not installed, so it failed from the first turn; but the same wedge would follow a transient claude fault mid-run. The outcome is unrecoverable without hand-editing runtime.json, and it burns tokens while stuck. Correcting my earlier Part B note: I had marked the backend-failure cancellation path as safe because it cannot cause a wrong or double resolution. That holds, but it can cause no resolution at all, a permanent wedge, which the earlier review missed.

Suggested fixes, in order:
1. A transient, retryable manager-backend fault must not cancel a pending human-approval gate. Separate "the manager LLM is briefly unavailable" from "this attempt is bad"; retry or park the turn and keep the command pending so the tools remain provisioned.
2. If a command is genuinely cancelled, actually abandon the attempt (stop relaunching workers) rather than leaving replacement_running looping, and cap the replacement loop.
3. On recovery, re-provision the resolution tools for a held request, or have the replacement worker re-establish a fresh pending command so _available_tool_names surfaces the tools again.

AndrewRqy:
Continuing with the manager and memory part (hitl_manager_react.py, hitl_manager_context.py, hitl_manager_history.py, hitl_manager_inbox.py, hitl_manager_host.py, hitl_world_model.py, hitl_manager_mcp.py). This part is in good shape. The two things I most wanted to check both hold up: the mutating-tool boundary is enforced at execute time (not just by which tools are listed), and the human-resolution binding cannot be steered by a stale browser. A few notes below, none severe, plus the duplication items.

1. The HITL web server has no auth, only the loopback bind (hitl_manager_host.py, hitl_web_server.py). The MCP bridge is loopback plus a bearer token plus a 403 on mismatch, but the web server relies entirely on binding to localhost. The container-mode escape hatch (NEURICO_HITL_WEB_CONTAINER_MODE=1 with host 0.0.0.0, host lines 401-407) trusts the operator to put a loopback-only Docker publish in front, which the code cannot enforce; a misconfigured publish would expose the manager surface, including conversation and resolution submission, unauthenticated. There is also no CSRF/Origin check, though the server-issued request keys blunt a blind cross-origin resolution. Suggestion: add the same optional shared-token gate the MCP bridge already uses, so the web server does not depend solely on the bind.

2. No bounds on inbox input or tool-result size. HitlManagerInbox.enqueue/update (inbox.py:72,102) cap neither message size nor queue length, and append_tool_result (context.py:181) has no size cap, so a single oversized final record is kept verbatim and can push the prompt past the model context window. Both are cheap caps; worth adding a per-message length limit and a queue-length limit on the inbox and a truncation cap on tool results.

3. Worker whiteboard content reaches manager context unbounded. _active_whiteboard_tips (world_model.py:289-310) strips but does not length-bound tip content before it is rendered into the manager prompt (runtime_digest at least slices to the last 8). Low severity since it is the lab's own worker and the digest is labeled as untrusted hints, but a per-tip length cap would be consistent with the 800-char cap the whiteboard already enforces on the add path.

4. _restore reads the conversation JSONL without the file lock in __init__ (context.py:68/74) while reload (86-90) takes it. Atomic os.replace on the write side makes a torn read unlikely, but the workspace-view read path constructs the context lock-free, and _restore raises fatally on a malformed line, so the read should take the same lock reload does.

Two naming notes, not bugs: runtime_digest is a prompt renderer, not a change-detector (re-sync is unconditional and idempotent, so there is no thrash), and design_panel does not spawn sub-agents despite the name and the PR description; it only updates ResearchState sections. Worth renaming or a comment so a future reader does not assume a sub-agent path exists.

Duplication:

- _now has five definitions in three formats, and one is a latent bug. research_state.py:80 uses naive local time while the idea log, runtime state, and manager context use UTC. Timestamps written into ResearchState will not order correctly against the authoritative stores. This should be one shared UTC helper; the format divergence is worth fixing regardless of consolidation.
- The loopback HTTP server plus bearer-token handler is implemented a third time here (the manager MCP bridge in hitl_manager_react.py:449-495), alongside the worker tool server in hitl.py and the MCP adapter in hitl_manager_mcp.py. The loopback-POST client is likewise a third copy. One shared server/client would also let the web server pick up the token gate from finding 1 for free.
- The idea-record schema is decoded in two places: the read-side projections in hitl_world_model.py (_idea_id, _idea_link, _premises, _decision_text, _decision_layer, 312-352) and the write-side validators in hitl.py. Not literal copies, but two encodings of the same record shape, so a schema change has to be made in both. A shared idea-record accessor would remove the drift risk.
- The channel subclasses override almost everything. HitlWebChannel and HitlTerminalChannel reimplement submit_input, poll_input, prompt, and emit/subscribe against the base ABC rather than extending it (HitlTerminalChannel does not even call super().__init__()). The inheritance is thin enough that composition would be clearer, and at minimum the deliberate dropping of history replay is worth a comment.
- Atomic tmp+fsync+os.replace is hand-rolled again in inbox and context; same shared-helper story as the earlier parts. File locking is correctly shared via hitl_lock, so that one is fine.

Overall this part is solid. The tool-gating and resolution-binding designs do what the PR says they do.

AndrewRqy:
Last part, the web and CLI frontend (hitl_web_server.py, hitl_workspace_view.py, cli/hitl_web.py, cli/hitl_workspace.py, and static/hitl/). This part is in good shape. The two things that could have been serious both come back clean, and the rest is UX feedback plus a few low-severity notes.

1. The CSP is present but permissive. hitl_web_server.py:82 sets script-src 'self' 'unsafe-inline', so the XSS safety rests entirely on esc() rather than on the CSP. Since hitl.js is external and the page has no inline scripts, dropping 'unsafe-inline' from script-src would make any future markdown() regression (for example adding a [text](url) rule without filtering javascript:) non-executable. Cheap defense in depth. style-src 'unsafe-inline' is still needed for the inline graph-node styles, so leave that.

2. No per-message or queue-depth cap on the inbox. do_POST caps the request body at 1 MB, but HitlManagerInbox.enqueue applies no length or count limit, and each enqueue does a full load and rewrite, so inbox.json can grow unboundedly with O(n²) cost. Same root as the inbox item in the manager review; one cap fixes both.

3. A few raw file reads escape the fail-closed wrapper. Most reads in HitlWorkspaceView convert errors to HitlWorkspaceViewError (which becomes a clean 409), but a few in _frontier/_activity, notably plan_path.read_text at hitl_workspace_view.py:123 after an is_file check, are outside that conversion, so a race-deleted artifact could escape as a 500 instead of a 409. Wrapping those matches the rest of the module.

4. Asset serving is safe by allowlist, not by containment. The exact-name match is what stops traversal today, which is fine, but a defensive asset.resolve().is_relative_to(ASSET_DIR.resolve()) would keep it safe if a future edit ever switches to a wildcard or extension-based content-type lookup.

UX and reliability:
- The full DOM rebuild on every 5s poll and every event drops focus and caret. A manager typing resolution feedback when a refresh lands loses their cursor mid-word. Preserving document.activeElement identity plus the selection range across render() would fix the most visible annoyance.
- refresh() is neither debounced nor serialized. The 5s interval, the four SSE listeners, and post-action refreshes all call it independently, so fetches can overlap and land out of order. A trailing debounce plus a request sequence number to drop stale responses would smooth this.
- Drawer internal scroll is not preserved across refresh, so reading a long proposal in the drawer jumps to top when the snapshot changes. The graph scroll is preserved, so the same capture/restore pattern applied to the drawer's scrollTop would do it.
- Minor: the main composer survives refresh but is lost on page reload (in-memory), while resolution drafts persist to sessionStorage. Making the two consistent would be nicer.

Duplication: the HITL web server is a trimmed fork of web_server.py, the SSE loop, the server scaffolding, the port-scan-on-bind, and the identical 1_000_000 body cap are all copied rather than shared. Factoring the common server/SSE plumbing into one place would remove the fork, and it would also let the web server pick up the token gate discussed in the manager review for free. The usual _now and atomic-write repeats apply here too.

after this i will start doing real deployment to see how neurico actually work out in this scenario and how we can improve it:)

Frankbest18:
Thank you for the detailed review. I went through the findings against the intended HITL authority model, implemented the defects that could violate that model, and applied narrower compatibility and deduplication changes where behavior was genuinely shared.

## Problems addressed

### Recovery and durable transitions

- I normalized attempt identifiers across frontier-decision creation, completion, and recovery.
- Frontier publication now resumes from every durable transition state instead of repeating or rolling back a finalized manager decision.
- Normal frontier transitions now reach `completed`, matching the recovery path.
- Initial-root publication is now durable across checkpoint creation, root initialization, mirroring, continuation configuration, and scoring-ref retirement.
- Resolved manager feedback is preserved across crashes until the replacement worker consumes it.
- Scoring-review identity is retained across restart instead of becoming a generic pending request.
- Final frontier pruning and selection now also occur after the last iteration.

### Scoring recovery and temporary refs

- I re-capture the reviewed workspace fingerprint after stale public scoring results are removed, including initial scoring.
- Temporary scoring refs are retired before older HITL state is restored.
- Ref retirement is idempotent, so replay remains safe after a crash.
- Reused attempt identifiers no longer collide with leaked refs.
- Resealing preserves optional stranded artifacts instead of deleting them.
- Scoring continues to use the candidate workspace interpreter so experiment dependencies remain available.
- Evaluator import resolution is isolated from unrelated candidate-controlled files under `scoring/`.

### Worker, manager, and runtime lifecycle

- Worker-process exit is no longer a stage-completion signal. It remains only a local liveness event.
- Candidate execution is joined to the manager scoring and frontier-decision lifecycle, so the outer loop cannot return a successful but empty candidate.
- Manager finalizers are selected from the runtime-owned request kind, distinguishing initial scoring from candidate frontier review.
- The outer pipeline can no longer reverse a manager-approved initial scoring assessment.
- The intended initial behavior remains unchanged: after manager approval of a valid score, runtime automatically publishes the initial experiment as the root.
- Manager-backend cancellation now terminates the affected attempt instead of repeatedly launching workers against an unresolvable request.
- A terminal finish-command latch replays finalized responses to delayed duplicate `hitl-finish-phase` calls.
- Worker instructions now require exactly one finish command and waiting for its response.

### Run ownership

- I replaced process-local ownership with a workspace-scoped cross-process lock.
- CLI and webpage launches now observe the same ownership state, preventing concurrent runs in one workspace.

### Rule-maker and evaluator contract

- Rule-maker outputs are structurally validated before sealing.
- The declared evaluator interface, evaluator, targets, and review log must be internally consistent.
- Manager review now checks that contract rather than relying on the worker’s completion claim.
- Early worker exit and incomplete rule-maker output recovery were strengthened.

### Manager context and history

- I added inbox message-size and queue-depth limits.
- Complete raw tool results remain archived, while bounded projections enter active context and searchable recall.
- Initial conversation restoration now uses the same lock as reload.
- Durable timestamps are normalized to UTC.
- Whiteboard projection reapplies the existing tip-length boundary.

### Web security and refresh behavior

- The HITL web server now has authenticated bootstrap, session cookies, and same-origin validation.
- `unsafe-inline` was removed from `script-src`.
- Artifact-read races produce controlled workspace conflicts instead of server errors.
- Refreshes are serialized and coalesced; stale responses are discarded.
- Focus, input selection, and drawer scroll survive refresh.
- The webpage cannot start another run while the workspace is owned by a CLI or another process.

### Ordinary NeuriCo compatibility

I also audited shared changes to keep ordinary AutoResearch behavior unchanged:

- restored the missing `@staticmethod` on ordinary proposal resolution;
- moved ordinary seal failures inside the existing attempt boundary;
- restored ordinary whiteboard writes to file-only behavior;
- kept strict sealing without deleting unrelated optional artifacts.

### Deduplication

I consolidated duplication where the semantics were genuinely identical:

- shared UTC and durable-write helpers;
- shared HITL paths, Git helpers, stage runtime, and AutoResearch primitives;
- one authoritative sealed-path and workspace-visibility policy;
- shared hashing and snapshot behavior;
- removal of unused ordinary Git-whiteboard and unseal paths.

I intentionally did not merge components merely because they have similar shapes. The ordinary and HITL web servers have different state, authentication, and lifecycle contracts. Worker and manager transports have different trust boundaries. Read-side idea projection is also not the same operation as write-side validation. Narrower plumbing can be shared later, but combining those components in this PR would broaden the change without strengthening its invariants.

## Design alignment

The implementation follows this authority model:

- Workers perform research and submit evidence, questions, proposals, and phase-completion requests.
- Runtime owns legal tool exposure, artifact and write-boundary validation, checkpointing, sealed scoring, durable transitions, and exact recovery replay.
- The manager owns semantic and scientific judgment, including repair and candidate acceptance or rejection.
- The human participates through explicit approval gates and manager `ask_human` requests.
- Runtime never uses provider exit, score magnitude, or generic phase status as a substitute for a required manager decision.

Two review suggestions were handled as design clarifications rather than applied literally:

1. HITL recovery intentionally has no arbitrary worker-attempt cap. Recoverable worker failure may continue until a durable decision is reached or the run is externally stopped. The defect I fixed was manager-backend cancellation repeatedly launching workers that could no longer resolve the cancelled request.

2. The sealed scorer intentionally uses the candidate workspace’s configured Python environment. This is required for experiment dependencies. I strengthened evaluator-file and import isolation instead of replacing it with a fixed NeuriCo interpreter.

AndrewRqy:
Thanks a lot for all these detailed fixes, really appreciate how much effort you put into this:) The fixes look great and after testing i think the issues i flagged out are properly handled. I wasnt that famuliar with front end development so if the relevant data structure looks good to you then we should stick to that. i agree with the design logic behind the worker retry logic, just think it may be good if we get an optional flag of worker failed attempts limit before the constant failure pattern is bring to the manager to discuss with human for alternative. The scorer one makes sense. Thanks again for the hard work, im merging this pr

AndrewRqy (approved):

PR #149, opened 2026-07-30
fix(hitl): strengthen phase completion handling

Frankbest18:
## Summary

This follow-up to [#146](https://github.com/ChicagoHAI/NeuriCo/pull/146) addresses reliability issues identified during post-merge HITL AutoResearch testing.

- Preserve ordered phase completion when a provider exits before its blocking runtime command returns.
- Reconnect replacement workers only when a durable held request genuinely requires replay.
- Treat the durable manager response as authoritative instead of using worker-process exit as a workflow-completion signal.
- Prevent client disconnects from turning an already-completed `/phase/finish` operation into a runtime error.
- Provide managers with the exact runtime-authorized MCP tool surface and required finalizer for each turn.
- Restrict Codex MCP access to the explicitly enabled tools and fail clearly when the required MCP server is unavailable.
- Require private evaluator inputs to be materialized under `data/.test/` and declared through `scoring/targets.json`.
- Validate declared evaluator inputs before sealing, including path containment, existence, regular-file type, non-empty content, symlink rejection, and undeclared-file detection.
- Align rule-maker, worker, revision, and manager-review instructions with the sealed-input contract.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@
 
 <hr>
 
-<!-- agent-summary: NeuriCo is an autonomous research framework. Input: YAML (title, domain, hypothesis). Output: code, results, plots, LaTeX paper, GitHub repo. Providers: Claude Code, Codex, Gemini. Install: curl one-liner or git+docker. Requirements: git + docker (or git + python 3.10+), plus an AI coding CLI (Claude Code, Codex, or Gemini). License: Apache 2.0. Repository: https://github.com/ChicagoHAI/neurico -->
+<!-- agent-summary: NeuriCo is an autonomous and human-in-the-loop research framework. Input: YAML (title, domain, hypothesis). Workflows: single-run research, scored AutoResearch, and manager-mediated HITL AutoResearch. Output: code, results, plots, LaTeX paper, GitHub repo. Providers: Claude Code, Codex, Gemini. Install: curl one-liner or git+docker. Requirements: git + docker (or git + python 3.10+), plus an AI coding CLI (Claude Code, Codex, or Gemini). License: Apache 2.0. Repository: https://github.com/ChicagoHAI/neurico -->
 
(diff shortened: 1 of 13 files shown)


AndrewRqy:
Hi! Thanks a lot for the hard work:) This pr looks clean and there i just one fragile point i want to mention:

The MCP server name neurico_hitl_manager is hardcoded in two independent places that must stay in sync — the config key (hitl_manager_react.py:563) and _mcp_allowed_tool_name (:502). The new Codex enabled_tools filter strips mcp__<configkey>__ from the allowlist that _mcp_allowed_tool_name builds, so it only works because both literals match.

If either is ever renamed without the other, startswith(allowed_prefix) matches nothing → enabled_tools=[] → Codex runs with zero tools → the manager can't call its finalizer → the turn wedges. It's silent (empty list is valid config, no error), and it would quietly reconstruct the exact wedge this PR fixes. The claude/gemini CLI path has the analogous failure via the allowlist.

Everything else looks great.

Frankbest18:
Thanks for catching the MCP namespace coupling. I addressed that specific issue by defining the manager MCP server name once as `_CLI_MCP_SERVER_NAME`. Both the generated MCP configuration key and the provider-facing `mcp__<server>__<tool>` allowlist are now derived from the same constant, so renaming the server cannot silently produce an empty tool list and wedge the manager turn.

Alongside that fix, this follow-up includes the additional HITL reliability fixes identified during testing:

- Prevents concurrent proposal submissions while another manager request is active.
- Preserves the correct request identity when scoring repair resumes a blocked worker phase.
- Prevents an already normalized scoring-repair response from being normalized a second time as a raw manager decision.
- Keeps terminal manager-provider cancellation from being converted into a misleading worker-repair loop.
- Tightens phase-finish and manager-finalizer tool exposure around the active runtime boundary.
- Improves durable runtime-state and frontier-transition handling across recovery.
- Aligns rule-maker and worker prompts with the single-phase completion contract, including avoiding duplicate or background phase-finish attempts.
- Keeps ordinary NeuriCo execution behavior outside the HITL-specific paths unchanged.

AndrewRqy:
Thanks a lot for the followup, everything looks great now, merging this pr

AndrewRqy (approved):

PR #151, opened 2026-07-30
Name repos mechanically from the idea, not via gpt-4o-mini

AndrewRqy:
## Problem
`_generate_repo_name` condensed the idea title into a slug via a `gpt-4o-mini` call (OpenRouter/OpenAI). When no `OPENROUTER_KEY`/`OPENAI_API_KEY` was set, or the call failed, it silently fell back to the raw idea id, producing names like `probabilistic_calibration_metr_20260730_075410_ac649b07`. This hit any run without an OpenAI-compatible key (for example claude-subscription runs), independent of the `--provider` agent.

## Change
Derive the slug deterministically from the idea.yaml content, tried in order **title -> hypothesis -> domain**, then the sanitized idea id as a last resort. Naming no longer depends on an external LLM or API key. Same `{slug}-{random}-{provider}` shape; apostrophes are stripped so possessives stay one word (`solver's` -> `solvers`).

`hypothesis` is threaded through `create_research_repo` and its two primary callers (`runner`, `submit`). `fetch_from_ideahub.py` is left to PR #147 (which is editing it).

## Testing
Ran the actual function against real workspace titles plus stress cases (blank/stopword-only title, punctuation/symbols, non-ASCII, all-empty). Every output is a valid GitHub name with a deterministic slug, and the fallback ordering resolves correctly.

### DIFF
--- a/src/cli/submit.py
+++ b/src/cli/submit.py
@@ -148,7 +148,8 @@ def main():
                         private=args.private,
                         domain=domain,
                         provider=args.provider,
-                        no_hash=args.no_hash
+                        no_hash=args.no_hash,
+                        hypothesis=idea.get('idea', {}).get('hypothesis', '')
                     )
 
                     github_repo_url = repo_info['repo_url']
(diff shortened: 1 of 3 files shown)


Chenxi1818 (commented):
It works on my end using the idea submitted on IdeaHub.

Image

Chenxi1818 (approved):
Thank you :)

PR #158, opened 2026-08-06
fix(hitl): scope the manager's --bare compaction key so it does not override your login

AndrewRqy:
The HITL manager's conversation-compaction step runs `claude --bare`, which by design skips keychain/OAuth and authenticates only via `ANTHROPIC_API_KEY`. For anyone signed in with `/login` or an OAuth token there is no way to satisfy it except a global `ANTHROPIC_API_KEY`, which then overrides that login for every call in the run and meters the whole session. Compaction only fires once the manager conversation grows past ~70% of its token budget, so the run works on the login up to that point and then a long HITL session dies partway through with a misleading `Not logged in · Please run /login`.

## What this adds

`NEURICO_HITL_MANAGER_API_KEY`: when set, the key is injected into only the `--bare` subprocess's environment. The main ReAct loop and the worker agents keep inheriting the parent environment, so they stay on your login/OAuth, and only the compaction call uses the scoped key. The global environment is never mutated, so nothing else switches to metered billing.

## Changes

- `src/interactive/llm_backend.py`: on the `--bare` path, pass `env={**os.environ, "ANTHROPIC_API_KEY": <manager key>}` to the subprocess when `NEURICO_HITL_MANAGER_API_KEY` is set, otherwise `env=None` (unchanged). When no key is resolvable, the existing `RuntimeError` is annotated with the real cause instead of the bare `Not logged in`.
- `tests/test_llm_backend_bare_auth.py`: the scoped key reaches only the `--bare` child, never the parent env or non-bare calls, and the failure hint is emitted.
- `.env.example`: documents the new optional variable.

With the variable unset, behavior is identical to today.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -14,6 +14,9 @@
 # ─── OPTIONAL: Additional API keys (passed through to agent environment) ───
 # Agents may use these during experiments (e.g., for model access, routing)
 # ANTHROPIC_API_KEY=
+# NEURICO_HITL_MANAGER_API_KEY=      # HITL only: Anthropic key for the manager's `claude --bare`
+#                                    # compaction step (it ignores login/OAuth); scoped to just that
+#                                    # call, so the rest of the run keeps using your login/OAuth.
 # GOOGLE_API_KEY=
 # OPENROUTER_KEY=
 # HF_TOKEN=                          # Hugging Face (private models/datasets)
--- a/src/interactive/llm_backend.py
+++ b/src/interactive/llm_backend.py
@@ -260,6 +260,7 @@ class LLMBackend:
         )
 
         # Build command
+        used_bare = False
         cmd = ["claude", "-p", "--verbose", "--output-format", "stream-json"]
         if self.model:
             cmd.extend(["--model", self.model])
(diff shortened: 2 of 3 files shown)


Frankbest18 (changes requested):
Thanks for investigating this issue and preparing the scoped-key approach. Since this PR was opened, #157 has addressed the same underlying problem directly.

The failure originated from `claude --bare`, which does not use OAuth or keychain authentication. #157 replaced it with:

```text
--safe-mode --tools ""
```

This keeps compaction isolated while preserving the user’s existing Claude authentication, so a separate API key is no longer required. It also reduced the number of compaction calls and improved provider error reporting.

Because #158 was based on the earlier implementation, it now conflicts with `main`. Resolving that conflict in favor of the `--bare` path could unintentionally restore the authentication problem fixed by #157.

I therefore suggest closing this PR as superseded by #157. The investigation was still helpful because it correctly identified `--bare` authentication as the source of the failure.

AndrewRqy:
ok great, i will close this pr then :)

PR #160, opened 2026-08-07
fix(hitl): exclude runtime-owned STATE.md from the write-boundary guard

AndrewRqy:
The runtime rewrites STATE.md (via `PipelineState._save()`) during guarded phases, but the public write-boundary guard snapshots it, so at phase finish the runtime's own write is flagged as a worker violation: `writes outside this HITL boundary: STATE.md`. Each rejection triggers a recovery, each recovery updates pipeline state and rewrites STATE.md again, so the phase can never pass. Observed live as 16 consecutive resource_finder recovery loops on a web-HITL run.

Fix: add `STATE.md` to the guard's excluded prefixes, like the other runtime-owned paths (`.neurico`, `logs`). Three unit tests pin the behavior: a mid-phase STATE.md rewrite no longer fails the plan gate, and genuine out-of-boundary writes are still caught.

### DIFF
--- a/src/core/hitl_workspace_guard.py
+++ b/src/core/hitl_workspace_guard.py
@@ -20,6 +20,9 @@ _EXCLUDED_PUBLIC_PREFIXES = {
     ".venv",
     "__pycache__",
     "logs",
+    # Runtime-owned: PipelineState._save() rewrites it during guarded phases,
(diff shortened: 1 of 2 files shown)


Frankbest18 (approved):
Thanks for fixing this!

PR #163, opened 2026-08-13
fix(hitl): improve web and terminal interface usability

Frankbest18:
## Summary

This PR improves the existing HITL web and terminal interfaces. The main goal is to make both interfaces show the same research state consistently, while leaving the underlying HITL workflow unchanged.

Previously, parts of the web and terminal interfaces derived status independently. That could produce stale stage labels, duplicate notifications, incomplete history, or different presentations of the same workspace. This update moves that interpretation into the shared workspace view. The web page and terminal client now render the same projected state in their own formats.

## Shared research status

`HitlWorkspaceView` now provides the common status used by both interfaces. It includes:

- the current research stage and phase;
- how long the current phase has been active;
- the next expected action;
- whether a run is active, paused, complete, or waiting for review;
- whether another run may be started;
- the pending human request;
- queued human messages;
- phase, idea, and review-resolution notifications.

The run launcher is still responsible for starting research, but it is no longer a separate source of status. Cross-process workspace ownership is also included, so a run started from one interface is visible from the other.

I added a bounded event record for interface-relevant transitions. It records phase changes, finalized ideas, and resolved reviews at existing runtime boundaries. These records are only for presentation. If recording one fails, the research workflow continues normally.

Repeated writes of the same phase no longer create duplicate notifications. Replacement processes and recovery work also remain internal unless the actual research phase changes.

## Web interface

The web conversation now combines manager messages and durable research notifications in chronological order. Notifications cover events such as:

- planning, execution, and review starting;
- scoring beginning;
- candidate decisions;
- frontier selection;
- paper writing;
- evidence, decisions, and proposals being recorded;
- human reviews being resolved.

Idea notifications remain concise in the conversation. The complete idea record is still available through the idea detail view.

The page now preserves conversation position when the user moves to the research view and returns. Drawer and graph scroll positions are preserved as well.

The composer was adjusted so that:

- it grows as the user types;
- it stops growing at a reasonable maximum height and then scrolls internally;
- multiline input is preserved when rendered in conversation;
- Enter sends the message;
- Shift+Enter inserts a new line;
- the conversation leaves enough space for the composer and queue instead of being covered by them.

The queue now behaves as an actual buffer. Submitting a message no longer immediately publishes it into conversation while another message is being processed. A queued message becomes part of the conversation only when NeuriCo claims it for processing. Claiming the message, recording it in conversation, and removing it from the queue are coordinated under the same inbox lock. This prevents deleted or unsent queue entries from appearing as user messages.

The queue has a maximum visible height and becomes scrollable when it grows beyond that height. Editing a queued message removes it from the queue and returns it to the composer. Sending it again adds the revised message back to the queue. When the queue is empty, a new message no longer visibly flashes through an unnecessary queue state.

The run control also uses the shared workspace status. It cannot start a second run while the workspace is already owned by another active process.

## Terminal interface

This PR improves the existing HITL terminal client; it does not introduce a separate terminal workflow.

The terminal now uses a small native input composer while leaving completed output in ordinary terminal scrollback. This avoids repainting the full screen and keeps resize and full-screen transitions from duplicating prompts or status rows.

When a workspace is reopened, the terminal replays:

- the complete human-visible conversation;
- durable phase notifications;
- recorded idea notifications;
- resolved human-review notifications;
- the currently pending request, when one exists.

The terminal does not print raw worker or provider output. Human messages and NeuriCo replies are rendered the same way whether they were produced during the current session or loaded from the existing workspace.

The terminal also adds:

- `/status` to show the current stage, phase, elapsed time, and next action;
- `/activity` to show recent durable research activity;
- `/idea <ID>` to show the complete record for a specific idea.

Conversation notifications use short summaries so routine activity remains readable. This does not truncate the authoritative idea record; `/idea <ID>` shows its complete content.

Terminal runtime output is written to the workspace log rather than mixed into the human conversation.

## Documentation

I updated the README, HITL AutoResearch guide, and workflow guide to document:

- `/status`
- `/activity`
- `/idea <ID>`


AndrewRqy:
Thanks a lot for the good work! Here are several points i will like to flag:

1. The conversation loop can die silently and wedge the whole interface.
In src/core/hitl_manager_host.py:1093, poll_input is called outside the try/except. poll_input now runs consume(publish), and publish writes the transcript (context append, history append, persist) while holding the inbox lock. If any of that raises (disk error, invalid queue entry), the exception escapes the loop and kills the conversation thread. The UI keeps accepting messages that queue forever with no error shown. Worse, the bad message stays at the head of the queue, so a restart hits it again immediately. Fix: move poll_input inside the try, or catch inside the loop, show a durable notice, and skip or quarantine the bad entry.

2. A client-supplied client_turn_id can make a processed message vanish from the transcript.
enqueue (hitl_manager_inbox.py:86-88) uses the client's client_turn_id as the record id, and publish passes it as record_id to transcript.append, which returns early if a record with that id already exists. Record ids are readable from /api/snapshot. So a request that reuses an existing record id gets its message delivered to the manager (chat(..., input_recorded=True), so the manager skips its own recording) but never written to the conversation. The durable transcript is the audit record for human oversight, so a message that runs without being recorded is a real integrity problem. Fix: always generate the id server side, or reject a supplied id that already exists or does not match the expected format.

3. Run failures are now invisible.
The launcher used to record failed plus the error string, and the web UI showed it. Now the launch thread logs the exception and nothing else, and _live_status can never return a failed state (the JS ships .workspace-state.failed styling that is now unreachable). A run that crashes at startup shows "Ready" or "Paused, progress is saved", which misleads the user. Also, the CLI configures no logging, so the log record goes to stderr and will corrupt the raw-mode composer rows. Fix: record the last launch error somewhere durable that live_status can read, and surface a failed state with the message.

4. Unknown key escapes insert garbage text into the composer.
native_terminal.py _escape_sequence knows a small set of sequences. For anything else it consumes only the ESC byte and the rest is inserted as literal text. PageUp inserts [5~, F1 inserts OP, Ctrl+Right inserts [1;5C. Fix: on an unknown ESC, swallow the complete CSI/SS3 sequence (\x1b[ up to the final byte, or \x1bO plus one char) instead of one byte.

5. Status polling does heavy file I/O 10 times per second.
The composer's readline loop calls the status callback every 0.1s select tick. Each call runs HitlWorkspaceView.live_status(): parse hitl_runtime_state.json (up to 500 events plus full pending commands), parse pipeline_state.json, and probe the run lock with a mkdir plus open plus flock. The CLI notification thread separately parses runtime state and the entire ideas JSONL every 0.5s. On local disk this is wasteful. On NFS (we run the DSI setup on /net/scratch2) it will be slow and chatty. Fix: cache the status with a minimum refresh interval (about 1s is enough for a timer display), or check file mtimes before re-parsing.

6. Strip control characters before printing worker content to the terminal.
Replay, notifications, /idea, and request bodies print idea and proposal text as is. That content is LLM-generated, often from fetched web pages, so it can contain ANSI escape sequences that move the cursor, clear the screen, or retitle the window. The web side escapes HTML correctly; the terminal side should strip C0/C1 control characters except newline and tab in HitlTerminalUI. This existed before, but this PR prints much more of this content.

7. Guard /status and /run against a corrupt state file.
hitl_manager_host.py:566 and the top of _launch_run_interactively call self._run_status() outside any try, unlike _live_snapshot which guards it. A corrupt or unreadable runtime-state file raises HitlWorkspaceViewError out of submit_input, and in interactive mode that exits the client.

Smaller things, fine as follow-ups: wide characters (CJK, emoji) count as one column, so the input row can wrap and break the two-row erase assumption; web submit_input no longer emits an SSE event on enqueue, so other open tabs only see a queued message on the 5s poll; present_run_status with force=False never renders anything, and stopping in hitl_cli.py is write-only now; a notification arriving while the thinking spinner is drawing can print onto the spinner line.

Frankbest18:
Thanks, Andrew. I went through each point and pushed a follow-up covering all of them.

1. A malformed queued message can no longer terminate the conversation loop or remain at the head of the queue indefinitely. Failed records are handled durably while later messages continue processing.

2. Client correlation IDs and durable transcript record IDs are now separate. A processed human message can no longer be omitted from the transcript by reusing an existing record ID.

3. Run startup failures are recorded durably and projected as a visible failed state, including the error message, instead of leaving the workspace looking ready or paused.

4. Unsupported terminal escape sequences are consumed completely rather than leaking fragments such as `[5~` or `OP` into user input.

5. CLI status projection is cached with a bounded refresh interval, avoiding repeated runtime, pipeline, idea-log, and ownership reads on every input-loop tick.

6. Stored terminal content is sanitized at the rendering boundary. Unsafe C0/C1 and ANSI control sequences are removed while preserving ordinary text, newlines, and tabs.

7. `/status` and `/run` now use the guarded workspace-state boundary, so malformed or temporarily unreadable runtime state is reported without terminating the CLI.

I also addressed the smaller follow-ups:

- Terminal layout now uses display-cell width for CJK characters, emoji, and combining characters.
- Web enqueue emits a workspace event so other open tabs receive queue updates immediately.
- The unused status signature and write-only stopping state were removed.
- Notification output now suspends and restores the thinking spinner so asynchronous updates do not overwrite each other.

Frankbest18:
One small follow-up fix: the CLI width handling imported `wcwidth` helper functions that are not part of its supported public API, which could prevent the HITL CLI from starting. I replaced those imports with local grapheme iteration while retaining `wcwidth.wcswidth()` for display-cell measurement.

This keeps the wide-character rendering fix intact and restores CLI compatibility with the project’s locked dependency environment.

AndrewRqy:
Thanks a lot for the detailed fix! Most concerns are addressed. I found three remaining edge cases:

  1. Existing queued records are not migrated. The previous schema was already version 2 and used client_turn_id as id, so a pre-update queued message can still collide with an existing transcript record and be processed without appearing in history. We should migrate records missing client_turn_id by preserving the old ID as client_turn_id and generating a new server-owned ID.

  2. Persistent inbox or transcript errors append a durable notice every 0.5 seconds indefinitely. This can flood the transcript while keeping later messages blocked. The error should be deduplicated/rate-limited, with whole-file corruption quarantined or surfaced as a single recoverable fault.

  3. CSI/SS3 handling looks good, but standalone Escape and Alt+UTF-8 remain problematic: Escape consumes the next typed character, and ESC + é inserts �. A short escape timeout and complete UTF-8 consumption should cover this.

  The other fixes look good :)

Frankbest18:
Thanks, Andrew. I addressed the two remaining edge cases that apply to the current implementation.

1. Persistent inbox or transcript failures no longer append a notice on every polling cycle. NeuriCo now emits one durable notice for a continuous failure episode and resets that state only after polling succeeds again. This preserves visibility without flooding the transcript.

2. Standalone Escape and Alt/Option input now use a short escape timeout. A lone Escape is handled independently, while an unsupported Alt sequence consumes only the Escape prefix and preserves the following complete UTF-8 character. Existing CSI and SS3 handling remains unchanged.

I did not add migration logic for the earlier queued-record shape. This interface has not been deployed with persisted user queues, so there is no supported legacy data to migrate. Keeping migration code for records that do not exist would add an unnecessary compatibility path.

I also included two queue fixes completed before this review:

- When the inbox is idle, the first submitted message is rendered directly as the active human turn. Only messages submitted behind an active turn appear in the editable queue.
- A workspace now permits only one active manager consumer. Additional interfaces may render the same durable workspace state, but they cannot concurrently consume the inbox or produce duplicate replies.

AndrewRqy:
Thanks for the hard work! The two requested fixes work, and the migration omission is reasonable if no persisted v2 queues were ever deployed. I found one remaining issue:

  - Medium — active-versus-queued rendering is inferred from stale browser state. A newly opened tab does not receive the current thinking event because subscriptions are not replayed. During an active turn, the inbox can also be empty because that message has already been claimed. Therefore /private/tmp/ neurico-pr163-review-25d0837/src/interactive/static/hitl/hitl.js:466 treats the next message as direct, then hides it from the editable queue even though it is actually queued.

Frankbest18:
@AndrewRqy Thanks for catching this. The stale-browser-state issue is now fixed at the ownership boundary rather than with another client-side heuristic.

The server now decides whether each accepted conversation submission is `direct` or `queued` while holding the same dispatch lock used to enqueue and claim messages. It tracks a claimed turn until that turn finishes and includes the pre-insert queue position in the decision. `/input` returns that authoritative disposition, and the browser only creates the transient direct-turn rendering when the server explicitly returns `direct`. A newly opened or refreshed tab can therefore have stale or absent event state without causing a message submitted during an active turn to disappear from the editable queue.

Since this review comment, the branch also includes two related follow-ups:

- Manager presentation state is retained in the host and included in `/api/snapshot`, with monotonically increasing sequence numbers shared with live status events. Refreshing or opening a renderer during an active manager turn now restores the thinking state instead of losing the animation.
- An active AutoResearch run now owns the manager backend choice. Starting a Claude or Codex run selects that backend for both workers and manager, the web selector is synchronized and disabled while the run is active, and server-side input handling rejects attempts to drift from the run provider. The selector becomes editable again when the run ends; Claude remains the idle default. Shared launch plumbing enforces the same provider invariant without changing the terminal UI.

AndrewRqy:
Thanks for the update! The is only one edge case remains:

  - Starting a Codex run while an earlier Claude manager turn is still thinking calls set_provider("codex") immediately in /private/tmp/neurico-pr163-review-
    1327113/src/cli/hitl_launcher.py:130.

  - The provider call runs outside _turn_lock in /private/tmp/neurico-pr163-review-1327113/src/core/hitl_manager_react.py:1945, so /private/tmp/neurico-pr163-
    review-1327113/src/core/hitl_manager_react.py:332 can replace the backend—and stop its MCP bridge—while the old Claude call remains active.

Frankbest18:
Thanks for catching this. 

I added a dedicated backend lifecycle lock around manager provider calls and provider replacement. A run requesting a different backend now waits for the in-flight manager call and its retries to finish. The existing MCP bridge remains available until that call returns, after which the bridge and backend are switched. 

Provider status uses a separate short-lived state lock, so web snapshots remain responsive while the launch is waiting.

AndrewRqy (approved):

PR #164, opened 2026-08-14
feat(hitl): Bring all NeuriCo ideas into one web interface

Frankbest18:
## Summary

This change is based on #163 and furtehr extends the HITL web interface from a single-workspace page into a project-level idea portal.

The portal presents the idea catalog in a persistent sidebar and opens the selected idea’s existing HITL workspace in the main view. Users can submit new ideas, inspect their original definitions, assign display names, reorder the catalog, and start or continue AutoResearch without restarting `hitl-web`.

## Idea catalog

Idea creation is driven by `ideas/schema.yaml` and submitted through the existing `IdeaManager` validation path. The portal does not maintain a second idea format.

Display names and ordering are presentation state. They are stored separately under `.neurico/hitl-web/catalog.json`, leaving submitted idea records, idea IDs, and workspace paths unchanged. Dragging is suspended while a display name is being edited so the rename field retains normal text-selection behavior.

## Workspace sessions

The portal starts one authenticated HTTP server and creates workspace sessions lazily as ideas are opened. Each session uses the existing `HitlManagerHost`, `HitlWebChannel`, `HitlWorkspaceView`, and `HitlRunController`; the portal only scopes their endpoints and events by idea ID.

Manager ownership remains per workspace. If an idea is already managed by the CLI or another web process, opening it in the portal returns an explicit ownership state instead of leaving the workspace on `Loading`. The rest of the catalog remains available, and the portal retries the workspace after the existing owner exits.

## Interface state

Portal snapshots now include the manager presentation status and effective provider. This preserves the thinking indicator across refreshes and keeps the composer backend synchronized with the selected workspace. During an active run, message submission uses the run-owned provider through the same resolver as the single-workspace web interface.

Manager execution failures are persisted as human-visible replies. The workspace projection also recognizes the small set of failure notices written by earlier versions before that metadata was added, so existing workspaces no longer show unanswered human messages for those turns.

Workspace load failures are rendered directly in the selected workspace, and disconnected pages keep the last successful projection visible. The disconnected indicator remains visible at narrow viewport widths even when the normal connection label is hidden.


AndrewRqy:
Thanks a lot for this great work:) There are three issues i wanna raise: 

 1. High — Docker cannot run portal-created or later-selected ideas with local resources. /private/tmp/neurico-pr164-review-9834c9b/docker/run.sh:2092 mounts resources only for the optional idea supplied when the container starts. Docker mounts cannot be added later, so resources for another selected idea—or a newly submitted idea—are unavailable and stage_local_resources() fails. This needs an architectural decision: mount a restricted shared resource root, or launch research in per-run containers with the required mounts.

  2. Medium — transient messages leak across workspaces. directTurn is global, /private/tmp/neurico-pr164-review-9834c9b/src/interactive/static/hitl/hitl.js:227 does not clear or scope it, and an in-flight response can set it after the user switches ideas. The message is then rendered in the wrong idea’s conversation and may remain indefinitely. Transient submission state should be keyed by idea ID, and async completions should verify the initiating idea is still selected.

  3. Medium — ownership is checked after workspace mutation. The portal constructs HitlManagerHost at /private/tmp/neurico-pr164-review-9834c9b/src/cli/hitl_web_portal.py:284, which initializes transcript/runtime/world-model files, while the consumer lease is only acquired later in /private/tmp/neurico-pr164-review-9834c9b/src/core/hitl_manager_host.py:1168. I reproduced an ownership rejection that still created SQLite, runtime, lock, and research-state files. The lease should be acquired before manager initialization.

There is also a scaling concern: every 10-second catalog refresh calls live_status() for every workspace at /private/tmp/neurico-pr164-review-9834c9b/src/cli/hitl_web_portal.py:197, including repeated YAML and runtime-state reads. This will become expensive on DSI/NFS as the catalog grows.

Frankbest18:
Thanks Andrew. I worked through the concerns you raised and did beyond those.

For Docker, the portal now hands launches back to the host through a durable request. The host claims each request once and starts a detached run container with the selected idea’s local resources mounted read-only. Because those mounts are resolved when the run starts, this also works for ideas selected or submitted after the portal container is already running. The handoff validates the idea, workspace, checkout, and resource manifest. Missing resources or container startup failures are recorded in the workspace instead of leaving the portal stuck in `Starting`.

The portal’s transient state is now scoped to the selected idea. Composer text, notices, request drafts, and pending provider changes cannot leak into another workspace. API responses and workspace events also verify that the initiating idea is still selected before updating the page. Conversation binding is deferred until the renderer lease has been acquired.

I removed the repeated project-wide `live_status()` scan as well. The idea catalog now reads metadata only. Live state comes from the selected workspace’s snapshot and event stream, so opening the all-ideas portal no longer performs expensive runtime inspection across every workspace.

The ownership model has been separated into renderer ownership and manager-consumer ownership. Web and CLI can render the same workspace at the same time, while one consumer processes its manager inbox. An active run owns that consumer. Manager messages remain active until their replies have been recorded successfully, and failed messages stay available for a later retry.

Runs now execute in a detached worker rather than inside the interface process. A run started from web or CLI continues after that interface closes, and a new renderer can reconnect to the same workspace. Web and CLI now read the same durable conversation, show messages submitted from either interface, preserve the thinking state across refreshes, and resolve the same HITL request without duplicating the decision or leaving a stale “already resolved” error.

Manager provider selection is now stored in the workspace runtime. Starting a run pins the manager to that run’s provider. Outside an active run, the workspace keeps the provider selected in the web interface, with Claude used as the default when no selection has been saved. Messages no longer store a provider themselves. The manager resolves the provider immediately before processing each attempt, and provider changes are serialized with active manager calls. If a provider attempt fails, the original message remains active and can be retried after a provider change, a new submission, or a manager restart.

AutoResearch continuation state now stores workspace-owned history paths portably. This allows the same workspace to continue when it is opened through local NeuriCo or Docker, where the absolute workspace path is different.

The CLI now has a `/stop` command. It asks for confirmation before stopping and restoring saved progress. The web stop button uses the same operation through a confirmation dialog, and Docker exposes it as `./neurico hitl-stop <idea_id>` with an optional `--yes` flag.

All three paths write the same launch-scoped stop request. The running worker observes it, terminates the active provider process, restores the latest rollbackable checkpoint, and records the run as stopped. The stop request is kept outside the rollback boundary so restoring the checkpoint cannot erase the request and restart the run. Proposal workers also check the stop state before creating replacements.

Finally, detached workers are reaped after they exit, regardless of whether the launch came from web or CLI. Completed launch records now preserve both start and completion times, and the workspace projection uses that completed record instead of stale pipeline timing fields.

AndrewRqy:
Thanks a lot for the fix! My previous issues have been addressed but there are couptle bugs found in the new lifecycle code:

- resolution-reply livelock that survives restarts. consume_resolution_reply (hitl_manager_host.py:308) calls the handler then complete_resolution_reply; submit_resolution_reply raises when no pending command exists (e.g. a rollback cleared it after submit). The raise skips complete_resolution_reply, so the durable reply is re-read every 0.5s forever, wedging the consumer — and it persists to every future consumer.
-  user stop misclassified as a stage failure. run_plan_centered_hitl_stage (hitl_stage_runtime.py:51-53) branches only on finish.get("approved"); a stopped result falls through to on_failed → runs the failure-rollback and records the stage as failed instead of a clean cooperative stop. 
-  submit_resolution_reply ignores the reply's own request_key (hitl_manager_react.py), applying a human answer to whatever command is currently pending — wrong command if it advanced.

Frankbest18:
I’ve updated the request lifecycle so rollback clears any saved reply for that request, and both CLI and web now preserve the request key when delivering a reply. Late replies are discarded instead of being carried into the restored state.

I also fixed the related stop behavior: /stop now propagates past the normal failure-recovery handlers, so the run restores its checkpoint and exits instead of resuming from it.

AndrewRqy:
Thanks a lot for the feedbacks! Here are some remaining small issues that i think is worth raising:

— consumer lease released before the conversation thread is confirmed dead (two-consumer race). _stop_manager_consumer (src/core/hitl_manager_host.py:1583-1584) joins the conversation thread with timeout=2, then releases the exclusive consumer lease regardless of whether the join succeeded. If manager.stop() doesn't unblock the turn within 2s (a turn blocked on an event or a non-cancellable backend call), the old thread is still alive when the lease is released, and the incoming handoff consumer (which waits up to 5s for the lease) acquires it. Two consumers then mutate the same inbox/transcript. The lease should not be released until the thread is confirmed stopped.

— a wedged _claimed_active_id permanently stalls a reused renderer. The channel's _claimed_active_id short-circuits poll_input to None while set (hitl_manager_host.py:280 web, :1248 terminal) and is cleared only in finish_active_turn. If the thread is force-joined by that 2s timeout without running its turn-finalizing path, _claimed_active_id stays set on the long-lived channel, so the next consumer on that channel returns None from poll_input forever and processes no inbox messages. _stop_manager_consumer resets the manager and lease but not the channel's claim.

— a run in the "starting" state cannot be stopped. request_hitl_run_stop (src/core/hitl_run_control.py:17-19) raises "No HITL AutoResearch run currently owns this workspace" when active_hitl_workspace_run returns None. During the Docker startup window (launch.json is starting, the worker hasn't claimed run.lock yet — the sleep 0.5/startup path in docker/run.sh:2221), a Stop is rejected and not persisted, so the intent is lost and the worker runs to completion. A stop requested before ownership is observed should bors it once it starts.

Everything else looks great, good work!

Frankbest18:
Manager ownership is now held until the existing conversation thread has fully exited. If a backend call takes longer than the handoff timeout, the handoff fails cleanly and the old consumer keeps its lease.

I also fixed the stale renderer claim. Web and CLI now clear their local claim even when durable settlement fails, so a replacement consumer can recover the message instead of leaving the renderer stuck.

Stop requests are now accepted while a launch is still starting. The detached worker checks for that request before entering the runner, and the runner checks again immediately after acquiring workspace ownership, before changing research state.

AndrewRqy (approved):

PR #167, opened 2026-08-18
feat(hitl): Add automatic manager-led research mode

Frankbest18:
## Summary

This PR introduces Auto as a second operating mode for HITL AutoResearch. This work is stacked on PR164, which has not been merged. The changes should be reviewed against `hitl-web-idea-submit`.

Full HITL and Auto use the same research pipeline, manager, workers, workspace history, scoring, checkpoints, recovery, and interfaces. The difference is who holds final decision authority during a run.

| Research boundary | Full HITL | Auto |
|---|---|---|
| Proposal admission | The manager reviews the proposal before it is presented to the user. The user decides whether the proposal may proceed. | The manager reviews the proposal and makes the admission decision. |
| Plan approval | The manager reviews the plan, followed by the required human approval. | The manager gives the final plan approval. |
| Worker-raised decisions | The manager may send the decision to the user and wait for a response. | The manager resolves the decision using the options provided by the worker. |
| Phase review | A phase can pause when human intent or approval is required. | The manager approves the phase or returns concrete revision instructions. |
| Human inbox | Human requests may be created and remain active until resolved. | Human requests are disabled for the duration of the run. |
| Decision records | Human decisions are recorded at A level. | Manager decisions are recorded at B level. |

Auto therefore preserves the HITL research structure while allowing a run to continue without waiting for user participation.

## Auto decision behavior

Auto does not treat manager review as a formality. The manager still evaluates the evidence at every existing HITL boundary.

A proposal can be rejected with feedback and returned to the proposer. A plan can be returned for revision. A completed phase can be sent back for repair. A successfully scored candidate can still be rejected when its result is weaker than the current frontier or introduces an unacceptable tradeoff.

Worker-raised decisions also remain explicit. The worker records the question, evidence, and available options through the existing HITL mechanism. In Auto, the manager must select from those options and persist its reasoning before the worker continues.

The runtime disables `ask_human` while Auto is active. This prevents an Auto run from entering a waiting state through a prompt mistake or an unexpected manager action.

## Switching between Auto and full HITL

The mode belongs to an individual run and is stored in durable workspace state. It is carried through launch, worker continuation, pending commands, recovery, stop handling, and final status.

A workspace can be started in Auto and later continued in full HITL. The reverse is also supported. Changing modes affects future and unresolved decisions; completed decisions remain unchanged in the workspace history.

Plan approvals retain their authority level. A manager-approved Auto plan is recognized when Auto resumes, while full HITL continues to require the human approval level expected by that workflow.

Launch records without a saved mode continue to use full HITL.

## Run configuration

CLI and web setup expose the mode as `Auto: Yes/No`, with Auto selected by default. The active workspace is labeled Auto or HITL so both renderers show the policy governing the current run.

The setting is part of research launch configuration. CLI and web remain interchangeable views of the same workspace.

Run setup can also be cancelled before launch. Provider, iteration count, paper selection, and paper style are validated before the launch request is created.

### DIFF
--- a/src/agents/autoresearch_proposer.py
+++ b/src/agents/autoresearch_proposer.py
@@ -164,7 +164,6 @@ def generate_autoresearch_proposal_prompt(
         hitl_stage="proposal",
         allow_raised_ideas=False,
         hitl_mode=hitl_mode,
-        human_resolution_allowed=hitl_mode != "auto",
         public_context=context,
         whiteboard_active_tips_md=whiteboard_active_tips_md,
         compute_backend_section=_generate_compute_backend_section(idea_spec, provider=provider),
--- a/src/agents/autoresearch_proposer.py
+++ b/src/agents/autoresearch_proposer.py
@@ -106,6 +106,7 @@ def generate_autoresearch_proposal_prompt(
     hitl_idea_reporting: bool = False,
     hitl_submission: bool = False,
     hitl_autoresearch_whiteboard: bool = False,
+    hitl_mode: str = "full",
 ) -> str:
     """
     Generate the AutoResearch proposer prompt from a curated public context.
@@ -162,6 +163,8 @@ def generate_autoresearch_proposal_prompt(
         pipeline_stage="experiment_runner",
         hitl_stage="proposal",
         allow_raised_ideas=False,
+        hitl_mode=hitl_mode,
+        human_resolution_allowed=hitl_mode != "auto",
         public_context=context,
         whiteboard_active_tips_md=whiteboard_active_tips_md,
         compute_backend_section=_generate_compute_backend_section(idea_spec, provider=provider),
@@ -273,6 +276,7 @@ def run_autoresearch_proposer(
         hitl_autoresearch_whiteboard=bool(
             env_extra and env_extra.get("NEURICO_HITL_AUTORESEARCH_WHITEBOARD") == "1"
         ),
+        hitl_mode=str((env_extra or {}).get("NEURICO_HITL_MODE", "full")),
     )
     prompt = append_prompt_block(prompt, prompt_suffix)
 
--- a/src/cli/hitl_launcher.py
+++ b/src/cli/hitl_launcher.py
@@ -93,10 +93,15 @@ class HitlRunController:
                 "Choose Claude or Codex for HITL research so the workers and manager "
                 "can use the same backend."
             )
+        raw_iterations = payload.get("iterations", 1)
         try:
-            iterations = int(payload.get("iterations", 1))
-        except (TypeError, ValueError) as exc:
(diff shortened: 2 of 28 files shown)


AndrewRqy:
Thanks for this, the Auto mode is well done. One blocking issue plus a few small ones:

High — switching Auto → Full deadlocks a pending plan phase-finish. adopt_hitl_mode (src/core/hitl_runtime_state.py) rewrites hitl_mode on the persisted pending command but never recomputes requires_human_approval. Scenario: start in Auto, a plan worker calls hitl-finish-phase (command persisted with requires_human_approval=False), the run stops before the manager finalizes, then it is continued in Full. On resume review_phase_finish's validator (from the current Full mode) demands ask_human, but _available_tool_names → _human_resolution_allowed_for reads the stale persisted requires_human_approval=False (hitl_manager_react.py:405-413, 446-447) and never offers ask_human. The manager can only finalize feedback, never approved, so the plan never clears the Full gate and the run is stuck. Asymmetric: Full → Auto is safe, only Auto → Full breaks. Fix: adopt_hitl_mode should recompute requires_human_approval for a pending phase_finish from the adopted mode and stage, or derive itmode instead of the persisted field.
                                                                                                                                                                Low — dead template variable. human_resolution_allowed=... is passe in several places (hitl.py:912/923/936/949,hitl_manager_react.py:1063/1177/1277) but no template reads it. Worth removing to keep the authoritative gate unambiguous.                                     
Low — dead command kinds. human_resolution_allowed (hitl_mode.py:42-44) accepts review_proposal/review_raised_idea/review_phase_finish, which are never emitted as command kind values.

Low — split validation. Auto proposal admission accepts a non-apprlidate and only rejects it in _finalize_proposal_manager_admission. Correct outcome, just fails later than it could.

Everything else checks out: ask_human is double-gated in Auto, B-level admission and raised-idea resolution persist correctly, requires_human_approval is coerced off, plan-approval A/B levels are retained across a switchty fails closed the same as Full. Once the switch bug is fixed this looks good to merge.

Also do you mind taking care of the conflicts with main as well :)

Frankbest18:
Thanks for the detailed review.

Resuming an unresolved plan review now recalculates requires_human_approval from the selected mode and stage. An Auto → Full switch restores human approval for a pending plan review and makes ask_human available, while execution-stage reviews remain manager-owned.

I removed the unused human_resolution_allowed prompt arguments and narrowed the authority helper to the command kinds that are actually persisted: proposal, raised_idea, and phase_finish.

Proposal admission validation now rejects inconsistent responses before finalization. Illegal proposals require concrete violations and feedback, legal proposals cannot include violations, and Auto feedback must provide concrete revision instructions.

AndrewRqy (approved):

PR #168, opened 2026-08-19
Port DSI iterations, eval verifier, and bootstrap baseline to HITL AutoResearch

AndrewRqy:
HITL AutoResearch (both Full and the Auto mode) was missing capabilities that Standard AutoResearch already had. This ports them onto the shared HITL pipeline. They are additive and mode-agnostic: they do not touch the decision-authority layer, so Full and Auto behave the same as before except where the new capability applies.

## DSI-Slurm remote compute in iterations

The HITL iteration comment handler ran every experiment locally (`dsi_remote_info=None`), so a HITL/Auto AutoResearch run could not use the cluster the way Standard AutoResearch iterations do. Each iteration's experiment now runs inside `dsi_slurm_remote_workspace(idea, work_dir)`, which provisions a per-attempt remote workspace and mounts the contract, and transient cluster artifacts are archived per attempt into the runtime-owned attempt directory before the checkpoint. This is a no-op unless `--compute-backend dsi-slurm` is selected, so local and modal runs are unchanged. `.neurico/dsi_slurm_remote_workspace.json` and `dsi-slurm-artifacts/` are also added to the checkpoint-exclude patterns as a safety net.

## Bootstrap an existing workspace into a frontier root

`--bootstrap-autoresearch-baseline` seeds the flat continuation state. The new `--hitl-bootstrap-autoresearch-baseline` runs the same mechanical bootstrap pipeline (workspace manifest, trimmer, bootstrap rule maker, scorer) against an existing unscored workspace, then publishes the scored result as the manager-driven AutoResearch frontier root. It needs no manager for the baseline itself. Continue it with the HITL web or CLI interfaces, or `--hitl-continue-autoresearch`, under the selected mode.

The bootstrap carries the same restart and recovery guarantees as the rest of the root AutoResearch workflow. Preparation and scoring both mutate the workspace before anything is published, so a durable pre-publication boundary is recorded in `HitlRuntimeState` (next to the initial-root publication transition, in the same style) before preparation begins. It holds the source checkpoint and a provider-local snapshot kept under the runtime-owned HITL state dir, which survives `restore_checkpoint`, `git clean -fd`, and process death. A later invocation detects an interrupted boundary and rolls it back, restoring both the public git workspace and the git-excluded provider dirs, before creating any new checkpoint, so a killed run is never adopted as a baseline. The boundary is retired only after the publication transition is durably established and owns recovery from there.

## Files

| File | Change |
|---|---|
| `src/core/hitl_autoresearch.py` | DSI remote workspace and per-attempt artifact archiving in the iteration comment handler; `construct_bootstrap_hitl_baseline` with the durable pre-publication recovery boundary and frontier-root publication. |
| `src/core/hitl_runtime_state.py` | Durable `bootstrap_prepublication_boundary` record (begin/read/clear), alongside the initial-root publication transition. |
| `src/core/autoresearch.py` | Exclude the dsi-slurm transient paths from checkpoints. |
| `src/core/runner.py` | `--hitl-bootstrap-autoresearch-baseline` entry point routed to the frontier bootstrap, added to the mutual-exclusion check. |

## Validation

The DSI wrapper resolves to a no-op for the local backend, so local and modal runs are unchanged. The bootstrap entry point is exposed on the CLI with a signature compatible with the flat bootstrap. Tests cover the recovery boundary end to end: the durable record surviving fresh instances, the interrupted-bootstrap rollback restoring both the public workspace and the provider dirs, the boundary being retired only after publication, and the failure edge cases (a failed rollback keeping its record and backup, the stored backup path never trusted as a delete target, and a crash between the two records reconciled on resume). Only the shared mechanism files are touched; none of the Auto/Full decision-authority code changes.

### DIFF
--- a/src/core/autoresearch.py
+++ b/src/core/autoresearch.py
@@ -75,6 +75,15 @@ PAPER_OUTPUT_PATTERNS = (
     "logs/paper_writer_prompt.txt",
     "logs/paper_writer_*.log",
 )
+# Transient dsi-slurm compute state: the remote-workspace contract file and any
+# pulled-back cluster artifacts. Both are normally cleared or archived before a
+# checkpoint; excluding them keeps a stale contract or transient artifacts out
+# of any checkpoint even if that cleanup did not run. Harmless when the compute
+# backend is not dsi-slurm (the paths simply never exist).
+DSI_SLURM_TRANSIENT_PATTERNS = (
+    ".neurico/dsi_slurm_remote_workspace.json",
+    "dsi-slurm-artifacts/",
+)
 
 CHECKPOINT_EXCLUDE_PATTERNS = (
     HIDDEN_SCORING_PATTERNS
@@ -84,6 +93,7 @@ CHECKPOINT_EXCLUDE_PATTERNS = (
     + AGENT_LOCAL_PATTERNS
     + PRIVATE_RUNTIME_PATTERNS
     + PAPER_OUTPUT_PATTERNS
+    + DSI_SLURM_TRANSIENT_PATTERNS
 )
 
 COMPARISON_EPS = 1e-6
--- a/src/core/hitl.py
+++ b/src/core/hitl.py
@@ -862,6 +862,10 @@ class HitlRuntime:
         self._tool_url: str = ""
         self._tool_token: str = ""
         self._tool_context: Dict[str, Any] = {}
+        # User-declared evaluation contract, surfaced to the manager as review
+        # criteria for the rule-maker scoring finish. Set per stage by the
+        # orchestrator; None when the idea declares no contract.
+        self._scoring_review_contract: Optional[Dict[str, Any]] = None
         self._phase_finish_result: Optional[Dict[str, Any]] = None
         self._phase_finish_request_key: str = ""
         self._phase_finish_response: Optional[Dict[str, Any]] = None
@@ -2338,6 +2342,17 @@ class HitlRuntime:
             },
         )
 
+    def set_scoring_review_contract(
+        self, contract: Optional[Dict[str, Any]]
+    ) -> None:
+        """Provide the user-declared evaluation contract for manager review.
+
+        Surfaced in the rule-maker phase-finish review so the manager can
+        confirm the scoring design honors the user's declared requirements. A
+        falsy contract clears it.
+        """
+        self._scoring_review_contract = contract or None
+
     def finish_tool_phase(self, payload: Dict[str, Any]) -> Dict[str, Any]:
         if not self._worker_request_lock.acquire(blocking=False):
             raise HitlActiveWorkerRequestError(self._pending_worker_command())
@@ -2818,7 +2833,7 @@ class HitlRuntime:
                     )
                     next_stage = "plan" if hitl_stage == "plan" else "review"
                     feedback = (
-                        "Runtime validation rejected this phase before manager review. "
+                        "Runtime validation found work outside the allowed HITL boundary. "
                         "Correct only these issues, preserve completed permitted work, then call "
                         "hitl-finish-phase again:\n"
                         + (issue_text or "- Recheck the active HITL workspace boundary.")
@@ -2976,6 +2991,7 @@ class HitlRuntime:
                 allow_scoring_approval=bool(self._tool_context.get("allow_scoring_approval"))
                 and hitl_stage in {"execution", "review"},
                 scoring_handoff_context=dict(self._tool_context.get("provenance") or {}),
+                declared_eval_contract=self._scoring_review_contract,
                 on_finalize=persist_phase_review,
                 on_scoring_approval=persist_scoring_approval,
                 hitl_mode=self.hitl_mode,
--- a/src/core/hitl.py
+++ b/src/core/hitl.py
@@ -2818,7 +2818,7 @@ class HitlRuntime:
                     )
                     next_stage = "plan" if hitl_stage == "plan" else "review"
                     feedback = (
-                        "Runtime validation found work outside the allowed HITL boundary. "
+                        "Runtime validation rejected this phase before manager review. "
                         "Correct only these issues, preserve completed permitted work, then call "
                         "hitl-finish-phase again:\n"
                         + (issue_text or "- Recheck the active HITL workspace boundary.")
--- a/src/core/hitl.py
+++ b/src/core/hitl.py
@@ -862,10 +862,6 @@ class HitlRuntime:
         self._tool_url: str = ""
         self._tool_token: str = ""
         self._tool_context: Dict[str, Any] = {}
-        # User-declared evaluation contract, surfaced to the manager as review
-        # criteria for the rule-maker scoring finish. Set per stage by the
-        # orchestrator; None when the idea declares no contract.
-        self._scoring_review_contract: Optional[Dict[str, Any]] = None
         self._phase_finish_result: Optional[Dict[str, Any]] = None
         self._phase_finish_request_key: str = ""
         self._phase_finish_response: Optional[Dict[str, Any]] = None
@@ -2342,17 +2338,6 @@ class HitlRuntime:
             },
         )
 
-    def set_scoring_review_contract(
-        self, contract: Optional[Dict[str, Any]]
-    ) -> None:
-        """Provide the user-declared evaluation contract for manager review.
-
-        Surfaced in the rule-maker phase-finish review so the manager can
-        confirm the scoring design honors the user's declared requirements. A
-        falsy contract clears it.
-        """
-        self._scoring_review_contract = contract or None
-
     def finish_tool_phase(self, payload: Dict[str, Any]) -> Dict[str, Any]:
         if not self._worker_request_lock.acquire(blocking=False):
             raise HitlActiveWorkerRequestError(self._pending_worker_command())
@@ -2991,7 +2976,6 @@ class HitlRuntime:
                 allow_scoring_approval=bool(self._tool_context.get("allow_scoring_approval"))
                 and hitl_stage in {"execution", "review"},
                 scoring_handoff_context=dict(self._tool_context.get("provenance") or {}),
-                declared_eval_contract=self._scoring_review_contract,
                 on_finalize=persist_phase_review,
                 on_scoring_approval=persist_scoring_approval,
                 hitl_mode=self.hitl_mode,
--- a/src/core/hitl_autoresearch.py
+++ b/src/core/hitl_autoresearch.py
@@ -14,8 +14,10 @@ import inspect
 import json
 import re
 import shutil
+import tempfile
 
 from core.autoresearch import (
+    AGENT_LOCAL_PATTERNS,
     AttemptHistoryManager,
     AutoResearchIterationResult,
     AutoResearchRunResult,
@@ -370,6 +372,40 @@ def run_fresh_hitl_autoresearch_initial_node(
     )
 
 
+# Provider-local directories that workspace preparation rewrites. Git
+# checkpoints exclude them (AGENT_LOCAL_PATTERNS), so a git restore cannot undo a
+# partial copy; the bootstrap snapshots and restores them alongside the public
+# checkpoint.
+_BOOTSTRAP_AGENT_LOCAL_DIRS = tuple(pattern.rstrip("/") for pattern in AGENT_LOCAL_PATTERNS)
+
+
+def _snapshot_bootstrap_agent_local(work_dir: Path, backup_root: Path) -> List[str]:
+    """Copy the provider-local dirs preparation mutates into a backup.
+
+    Returns the names that existed, so restore can recreate exactly the prior
+    state (a dir absent before preparation is removed on restore).
+    """
+    existed: List[str] = []
+    for name in _BOOTSTRAP_AGENT_LOCAL_DIRS:
+        source = Path(work_dir) / name
+        if source.exists():
+            shutil.copytree(source, Path(backup_root) / name, symlinks=True)
+            existed.append(name)
+    return existed
+
+
+def _restore_bootstrap_agent_local(
+    work_dir: Path, backup_root: Path, existed: List[str]
+) -> None:
+    """Restore the provider-local dirs from the pre-preparation snapshot."""
+    for name in _BOOTSTRAP_AGENT_LOCAL_DIRS:
+        target = Path(work_dir) / name
+        if target.exists():
+            shutil.rmtree(target, ignore_errors=True)
+        if name in existed:
+            shutil.copytree(Path(backup_root) / name, target, symlinks=True)
+
+
 def construct_bootstrap_hitl_baseline(
     *,
     idea: Dict[str, Any],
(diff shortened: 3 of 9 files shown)


Frankbest18 (changes requested):
Thanks for bringing these missing capabilities into the HITL path. I think this needs another pass before merging, mainly because the eval-verifier integration does not follow the revision flow used elsewhere in HITL.

**[P1] Eval verification should happen before manager review**

The verifier currently runs after the manager has approved the rule-maker output. If it rejects the contract, the stage is rolled back and then relaunched by the outer HITL loop. Because the verifier’s findings are not given to the next worker, this can repeatedly regenerate and reject the same contract while making additional worker, manager, and verifier calls.

More importantly, rollback is not the right mechanism for this situation. A rejected contract is an expected, recoverable workflow result: NeuriCo knows what failed and has concrete violations that the worker can address. Rolling back discards that useful context and turns an informed revision opportunity into a blind restart.

I think verification belongs in the validation flow before manager review. If verification fails, the current work should be preserved and the violations returned to the worker for revision. Only a contract that passes verification should reach the manager.

This would keep the behavior consistent with the rest of HITL: validation failures lead to informed revision, while rollback is reserved for unexpected external or systemic failures where NeuriCo can no longer safely continue from the current state. Verifier crashes or timeouts should also be distinguished from an actual contract rejection rather than handled as the same kind of failure.

**[P2] Interrupted root publication can be reported as complete**

The bootstrap path returns success as soon as it finds an existing frontier file. However, that file is created partway through the publication process, before continuation metadata, history mirroring, cleanup, and the final completion state have necessarily been written.

If the process stops after the root is initialized but before those later steps finish, the next bootstrap run sees the frontier and reports success without completing the pending publication. The workspace can therefore appear successfully initialized while still missing information such as the history root, lineage source, or last iteration. A later HITL continuation may then fail because the frontier is not actually ready to continue from.

**[P2] A workspace-preparation failure can leave partial changes behind**

The original workspace checkpoint is created before preparation begins, but `prepare_workspace()` runs outside the error-handling section that restores that checkpoint.

Preparation modifies several parts of the workspace, including provider skill directories and `.gitignore`. If one of those operations fails after earlier changes have already succeeded, the bootstrap command reports failure without restoring the original workspace. The user is left with a partially prepared workspace even though the bootstrap itself did not complete.

I would also appreciate focused tests for these paths, particularly verifier rejection, interrupted root publication, and failure during workspace preparation.

AndrewRqy:
Thanks for the detailed review! All three should be addressed.

[P1] Eval verification moved into the validation flow. The verifier now runs in the rule-maker phase-finish validator, before manager review, instead of after approval. A rejected contract is returned to the worker as an informed revision with the concrete violations, and completed work is preserved (no rollback, no blind restart). Only a contract that passes verification reaches the manager. Verifier crashes and timeouts are handled distinctly from a rejection: they surface as an infrastructure issue that tells the worker not to rewrite the contract and to retry, rather than fabricating violations. It runs on every phase finish whose artifacts validate, not just review, since the manager can approve a non-plan finish at any phase.

[P2a] Interrupted publication no longer reported complete. The pending-publication resume now runs before the HitlFrontierStore.exists() check. Since the frontier file is written midway through publication (at root init,ata, mirroring, and cleanup), the old ordering could report ahalf-published root as done. Resuming the transition first is replay-safe and idempotent. 
                                                          
[P2b] Preparation failure now restores the workspace. Moved prepare_workspace() inside the try that restores the source checkpoint, so a mid-preparation failure (skill dirs, .gitignore) rolls back instead of leaving a p

Tests. Added tests/test_hitl_autoresearch_review_fixes.py coveringier-error-vs-rejection, interrupted-publication resume, andpreparation-failure restore. Full suite green.

Frankbest18 (changes requested):
[P1] A model-driven verifier now controls HITL progression without HITL recovery guarantees

Thanks for making the changes from my earlier review. Moving verification before manager review is better than rolling the stage back. However, after looking more closely at how this verifier works, I think the concern is more fundamental than where it runs.

The verifier is another model-driven agent, not a deterministic runtime check. It may time out, stop early, produce malformed output, or make an incorrect judgment. Its verdict nevertheless decides whether the rule-maker’s work is allowed to reach the manager. This gives an unverified agent result direct control over the workflow, even though the verifier does not have the supervision that we built around normal HITL workers.

The verifier is not managed as a normal HITL stage: there is no durable request, continuation, replacement worker, or feedback loop for correcting its output. If it times out or fails to produce a verdict, that failure is returned to the rule-maker, even though the rule-maker cannot fix it. Retrying the same `hitl-finish-phase` request may also return the cached response without running the verifier again. A malformed but parseable verdict can be treated as a genuine contract rejection, causing the verifier’s own mistake to be sent to the rule-maker as feedback.

I am also not convinced that a separate reviewer is needed here. The manager already reviews the scoring design, including the metrics, targets, evaluation split, and required artifacts. Running another model to review the same work, and then allowing only work approved by that model to reach the manager, adds another failure point without creating a clear new responsibility.

Before changing this integration further, please compare it with the existing HITL design in `docs/HITL_AUTORESEARCH.md`, especially the sections on authority, stage lifecycle, scoring, recovery, and failure behavior. The relevant implementation is mainly in `src/core/hitl.py`, `src/core/hitl_stage_runtime.py`, `src/core/hitl_manager_react.py`, `src/core/hitl_runtime_state.py`, and `src/core/hitl_autoresearch.py`. These paths show why semantic decisions are manager-owned, why mandatory workers use durable continuation and replacement, and why rollback is kept separate from ordinary revision. A new workflow stage needs to preserve the same guarantees.

My preferred direction is to keep deterministic checks in runtime validation and move the semantic verification criteria into the manager review. If we still want to run the verifier, its report could be given to the manager as additional evidence, but it should not directly control the workflow.

If the verifier must remain a blocking stage, it needs to be treated as a proper HITL worker with the same durable continuation, replacement, and recovery support as the other stages. In its current form, a failure or bad judgment from this single agent can prevent the rule-maker stage from making progress and weaken the reliability guarantees of the HITL workflow.

AndrewRqy:
Thanks for this feedback. After reading more into the design i agree that the role of the manager already cover what the verifier should do, so not letting verifier do deterministic judgement here should be the right call. Based on the importance of matrix formation, i think more caution cannot hurt since as you said above it is possible for llm agents to make incorrect judgement, so i initially put the verifier here as a second gate to make sure the rules defined are well-formed. An alternative design here is that we can invoke the verifier to check on the product and provide an estimation to the manager for reference and let manager makes the call based on both models' judgements. But we can discuss the necessity of this kind of adversarial design later and i will remove the verifier part from the pr for now :)

Frankbest18 (changes requested):
Thanks, Andrew. This is definitely closer, and removing the verifier as a gate resolves the authority problem from my previous review. I still see two issues in the updated implementation.

**[P1] Contract verification is now assigned to an actor that must not inspect the evaluator**

The manager prompt now asks the manager to confirm that the declared metrics, targets, split, and required functions are correctly implemented in the evaluator. However, `scoring/eval.py`, `scoring/targets.json`, and `scoring/rule_maker_log.md` are protected by `SEALED_PATHS`, and the manager’s workspace inspector refuses to read them.

That restriction is intentional and should remain. The manager persists into later experiment stages, so exposing the hidden evaluator implementation to it could leak information into later worker feedback. We should not solve this by giving the manager access to those files.

As written, the manager receives the user’s contract but not the implementation evidence needed to verify it. It can only rely on the rule maker’s plan and finish summary. Runtime validation will catch syntax, ABI, and structural problems, but it will not catch a rule maker that reimplements a required function, ignores its output, or mistranscribes a metric in `targets.json`.

This suggests that an isolated verifier may still have a useful role, but only as an evidence producer. It could inspect the hidden implementation and provide a sanitized conformance report for the manager to consider. It should not expose the evaluator contents to the manager, and its verdict should not directly approve, reject, or otherwise control the HITL workflow. Failures in that verifier should also be handled as failures of the verifier itself, not returned to the rule maker.

**[P2] The bootstrap rollback boundary still ends before the durable publication handoff**

The rollback-protected section now correctly includes workspace preparation and the scoring pipeline. However, it ends immediately after `run_pipeline()` succeeds.

The code still needs to read the experiment plan, resolve and encode the history root, and call `begin_initial_root_publication_transition()`. Those operations happen outside the rollback handler. If one of them fails, the workspace has already been modified and scored, but the original checkpoint is not restored and no publication transition exists for the next run to resume.

This leaves the workspace between two recovery boundaries: it is no longer the original workspace, but it has not yet entered the replayable publication transaction. A later bootstrap run can therefore treat this partially bootstrapped workspace as its new starting point.

The original checkpoint needs to remain the recovery boundary until the publication transition has been durably recorded. Once that transition exists, publication should continue to use the existing replay-forward behavior rather than rollback.

AndrewRqy:
Thanks for the feedbacks, both are addressed.

[P1]  I think a verifier agent that generates only report can be useful in this case, i will make it in a followup pr

[P2] Fixed. The original checkpoint now stays the recovery boundary until begin_initial_root_publication_transition() records the transition durably. Preparation, scoring, the plan read, history-root resolution, and the transition record all run inside the restore boundary, so a failure in any of them restores the original workspace instead of leaving a scored-but-unpublished one for a later run to adopt. Only _commit_initial_root_publication stays outside, using the existing replay-forward/resume behavior. Added a test that a begin_transition failure restores and never commits.

Frankbest18 (changes requested):
Thanks, Andrew. The publication side now looks consistent with the existing replay-forward design, but I think there is still one recovery gap.

**[P2] Preparation is outside the effective scope of the rollback mechanism**

`restore_source()` uses `CheckpointManager.restore_checkpoint()`, which restores the public Git workspace. However, preparation also deletes and replaces files under `.claude/`, `.gemini/`, and `.codex/`. Those directories are excluded from checkpoints by `AGENT_LOCAL_PATTERNS`, and `git clean -fd` preserves ignored paths.

This means a failure halfway through `_copy_workspace_resources()` can still leave partially copied or overwritten provider skills behind. The public workspace is restored, but the complete pre-bootstrap state is not. The current test uses a fake checkpoint manager and confirms only that `restore_checkpoint()` was called, so it does not catch this difference. With the real `CheckpointManager`, I can reproduce a file under `.codex/skills/` surviving the rollback.

The relevant existing recovery design is in `src/core/hitl_stage_runtime.py`, especially `HitlStageRollback`, together with `HitlGitStateStore` in `src/core/hitl_git_state.py`. These show the two-layer boundary used elsewhere in HITL: the public workspace and runtime-owned state are captured and restored together. The bootstrap path may not be able to use `HitlStageRollback` directly because it is not running as an ordinary interactive stage, but it should preserve the same guarantee for every class of state that preparation mutates.

It would also be useful for the test to use the real checkpoint behavior with a pre-existing provider skill directory and verify the actual filesystem after a preparation failure.

AndrewRqy:
fixed. Preparation rewrites .claude/.gemini/.codex/skills/, which are excluded from git checkpoints by AGENT_LOCAL_PATTERNS and preserved by git clean -fd, so restoring the public checkpoint could not undo a partial copy. The bootstrap now snapshots those provider-local dirs before preparation runs and restores them alongside the public checkpoint, so a failure anywhere in preparation reaches the complete pre-bootstrap state. That covers both classes of state preparation mutates: the provider skills through the new snapshot, and .gitignore which git already tracks. The snapshot is discarded on every exit.

The test now reproduces the reported case, a file under .codex/skills/ surviving rollback, and verifies the real filesystem after a preparation failure, with the public checkpoint restore faked but the provider-skill restore real. It also covers the snapshot/restore round-trip and the case where a dir absent before preparation is removed on restore.

Frankbest18 (changes requested):
Thanks, Andrew. This fixes the provider-skill restoration when bootstrap fails through a normal caught exception. However, I think the recovery boundary is still incomplete in a more important way.

**[P1] The pre-publication rollback boundary does not survive process interruption**

The provider-local snapshot is created in a temporary directory, while both its location and the source checkpoint SHA exist only in the current Python process. Recovery depends on reaching the local `except` and `finally` blocks.

If the process is terminated during preparation or scoring, neither block runs. On restart, there is no durable rollback record and no initial-root publication transition yet. Bootstrap can therefore create a new source checkpoint from the partially prepared or scored workspace and treat that state as the new baseline. The temporary backup may still exist on disk, but nothing records where it is or that recovery is pending.

This is the kind of external or systemic failure for which rollback is intended. Handling an ordinary exception is useful, but it is not enough for a recovery boundary in the AutoResearch workflow.

There is also a smaller issue in `_restore_bootstrap_agent_local()`: `shutil.rmtree(..., ignore_errors=True)` can suppress a failed removal. When a provider directory did not exist in the snapshot, there is nothing to copy afterward, so the function can return successfully while partial state remains.

The publication side of this implementation already follows the right model: the transition is recorded durably, and a later process resumes it before making any new decision. I think the pre-publication side needs the same property.

Good references are:

- `src/core/hitl_stage_runtime.py`, especially `HitlStageRollback`
- `src/core/hitl_git_state.py`, especially `HitlGitStateStore`
- `src/core/hitl_runtime_state.py`, for durable transition records
- The recovery, stage-lifecycle, and failure-behavior sections of `docs/HITL_AUTORESEARCH.md`

Bootstrap may not be able to use `HitlStageRollback` unchanged because it runs outside an ordinary interactive HITL stage, but it should reuse or adapt the same public-plus-private recovery design. Before preparation begins, the source checkpoint and provider-local snapshot need to be recorded durably. A later invocation should detect and restore an unfinished bootstrap boundary before it can create another checkpoint or adopt the current workspace. That recovery record should only be retired after the initial-root publication transition has been durably established.

I would avoid adding another process-local backup path here. This is part of the root AutoResearch workflow, so it should carry the same restart and recovery guarantees as the rest of HITL.

AndrewRqy:
Thanks for the heads-up! Sorry i wasn't that familiar with the publication update logic before, I made the adjustments to make the backup procedure mimic what we do for workspace succession.  I added a bootstrap_prepublication_boundary record to HitlRuntimeState, next to initial_root_publication_transition and in the same style. It is recorded before preparation changes anything and carries the source checkpoint plus the provider-local snapshot location. The snapshot moved out of /tmp into .neurico/hitl/bootstrap_agent_local_backup/, the runtime-owned state dir that both restore_checkpoint and git clean -fd preserve, so it survives process death. A later invocation detects a pending boundary and rolls it back before creating any new checkpoint, restoring both layers (the public git workspace from the checkpoint, and the git-excluded provider dirs from the snapshot), so a killed run is never adopted as a baseline. The boundary is retired only after begin_initial_root_publication_transition() is durably established, at which point the publication transition's replay-forward resume owns recovery. The two records never both claim the run.

I do not reuse HitlStageRollback directly since bootstrap is not an interactive stage with a worker session, but it adapts the same public-plus-private capture-and-restore guarantee, resumed on restart.

Smaller issue. _restore_bootstrap_agent_local no longer uses rmtree(ignore_errors=True), so a failed removal surfaces as an incomplete rollback instead of returning successfully with partial state.

Tests cover the durable record round-trip across fresh instances (proving it survives process death), idempotency, the resume path rolling back an interrupted boundary, the record being written before preparation and retired only after publication, and the removal-failure surfacing.

Frankbest18 (changes requested):
Thanks for the great work, Andrew. The durable pre-publication record addresses the main restart concern. I think there are still three edge cases we should look at.

**[P1] A failed rollback removes its own backup**

If `_restore_bootstrap_agent_local()` fails, the pre-publication boundary stays pending, but the outer `finally` still deletes `bootstrap_agent_local_backup`.

On the next run, NeuriCo finds the pending boundary but the backup is gone. It skips restoring the provider directories and clears the boundary. The workspace can then be left partially restored with no backup or recovery record remaining.

**[P1] The backup path is trusted as a deletion target**

`agent_local_backup` comes from the saved runtime record and is later passed directly to `shutil.rmtree()`.

The only validation is that the string is non-empty. There is no check that it actually points to this workspace’s bootstrap backup. If the runtime record is damaged or changed, recovery could read from and recursively delete a different directory.

**[P2] A crash can leave both recovery records active**

Creating the publication transition and clearing the pre-publication boundary are two separate state writes.

If the process stops between them, both records remain. The next run finishes publication, but returns without clearing the older boundary or its backup. The workspace is then successfully published while still carrying a stale record saying that the bootstrap should be rolled back.

The closest existing example is `_recover_experiment_runner_from_runtime_checkpoint()` in `src/core/pipeline_orchestrator.py`. `HitlStageRollback` in `src/core/hitl_stage_runtime.py` and `HitlGitStateStore` in `src/core/hitl_git_state.py` are also useful references for how recovery state and its backing snapshot are kept together.

One small cleanup item: the PR description still says the eval verifier is being added as a hard gate, although that part has been removed.

AndrewRqy:
Thanks, all three are addressed, and I lined them up with the recovery style in _recover_experiment_runner_from_runtime_checkpoint and HitlStageRollback.

[P1a] The unconditional finally that deleted the backup is gone. The backup is now deleted only by a single retirement step that clears the record and deletes the backup together, reached only after a clean restore, the same way HitlStageRollback.restore() discards its snapshot last. A failed restore keeps both the record and the backup, so the next run retries instead of finding a pending boundary with no backup.

[P1b] Recovery no longer trusts the stored path as a delete target. It derives the one canonical backup location from the workspace and validates the stored path against it, raising on mismatch, the same way the experiment-runner recovery validates its refs/neurico/hitl-rollback/ ref. A missing source checkpoint and a failed provider-local restore also raise with context now, rather than being silently skipped or surfacing a raw OSError.

[P2] Since recording the publication transition and clearing the boundary are two writes, a crash between them left both active. The publication-resume path now retires the obsolete boundary and its backup, because once a publication transition exists the pre-publication window is over and publication owns recovery. A published root no longer carries a roll-back record.

Also refreshed the PR description, which still described the removed eval-verifier gate.

Frankbest18 (changes requested):
Thanks Andrew. I found one remaining problem in the recovery path:

[P1] A missing provider-state backup is treated as a successful rollback

In _rollback_bootstrap_prepublication_boundary, the .claude, .gemini, and .codex state is restored only when the backup directory still exists. If that directory is missing, NeuriCo restores the public checkpoint, skips the provider-local restoration, and then clears the recovery boundary.

That leaves any partial workspace preparation in place, while also removing the record that tells NeuriCo the rollback was incomplete. A later attempt can therefore adopt that partial provider state as its starting point.

The existing experiment-runner recovery path in _recover_experiment_runner_from_runtime_checkpoint and HitlGitStateStore is a useful reference here: a missing private snapshot is treated as an incomplete recovery, and the recovery boundary is not cleared.

AndrewRqy:
Thanks a lot for the check, this issue should be fixed. The rollback restored provider-local state only when the snapshot dir existed but cleared the boundary either way, so a missing snapshot left partial state in place while erasing the record that recovery was incomplete.

Following the missing-private-snapshot handling in _recover_experiment_runner_from_runtime_checkpoint, a missing snapshot now raises and keeps the boundary for the next run to retry, instead of clearing recovery over partial state. The only case that still succeeds is when no provider dirs existed originally, since restoration there is just a removal.

Frankbest18 (approved):

PR #170, opened 2026-08-20
fix: sanitize staged files before GitHub upload

### LINKED ISSUE #153: [Security/Feature] Sanitize sensitive API keys/tokens from logs before any GitHub upload
## Problem Statement

This issue was observed during today's run:
`./neurico fetch https://hypogenic.ai/ideahub/idea/dN5DnL4Gnuz5jGSCEa64 --submit --run --provider claude`
Claude flagged a potentially hazardous implementation and aborted the run early (unlike Codex, which completed successfully). During this automated run, sensitive secrets like OpenAI API keys or GitHub Personal Access Tokens (PATs) can be logged or captured in output/dump files, like today's `src/core.*` file. Unsanitized log files containing active credentials were pushed to the remote repository, triggering GitHub and provider automated security scanners to revoke the keys immediately.

Currently, neurico implements a key-sanitization mechanism intended to scan for key-like strings and delete them after execution. However, this cleanup step appears to be bypassed or aborted when the execution fails.

Since log files serve as necessary execution archives and cannot simply be excluded via `.gitignore`, we need a robust, fault-tolerant sanitization mechanism integrated directly into all GitHub upload workflows.

mimosapudical:
## Summary

- sanitize staged textual artifacts at the GitHub commit boundary
- re-stage sanitized files so stale secrets cannot remain in the Git index
- verify staged blobs before commit and fail closed if recognized credentials remain
- preserve binary files and safely handle staged deletions

Closes #153.

## Dependency

Depends on #169.

The first commit in this branch (`90d67f5`) is already under review in #169.
The new change introduced by this PR is `6fe92e0`.

Once #169 is merged, I will rebase this branch onto the updated `main` so this PR contains only the #153 change.

## Why

Previously, sanitization only covered selected files under `logs/` before `git add`.
Secrets written to other artifacts could therefore enter the Git index.

Also, sanitizing only the working-tree file after staging is insufficient because the Git index may still retain the original secret-containing blob.

The new boundary is:

agent artifacts
→ git staging
→ staged-file sanitization
→ re-stage sanitized files
→ staged-content verification
→ commit
→ push

## Testing

- `python -m pytest -q tests/test_security.py` — 13 passed
- `python -m py_compile src/core/security.py src/core/github_manager.py tests/test_security.py` — passed
- manual local Git acceptance test — fake secret absent from working tree, Git index, and committed blob

Full `tests/`: 178 passed, 3 unrelated existing Windows path assertion failures.

### DIFF
--- a/src/core/github_manager.py
+++ b/src/core/github_manager.py
@@ -15,7 +15,12 @@ import subprocess
 import shlex
 from datetime import datetime
 
-from core.security import sanitize_logs_directory
+from core.security import (
+    SanitizationError,
+    contains_sensitive_data,
+    sanitize_file,
+    sanitize_logs_directory,
+)
 
 try:
     from github import Github, GithubException, Auth
@@ -289,6 +294,11 @@ class GitHubManager:
             # Add all files
             repo.git.add(A=True)
 
+            sanitized_files = self._sanitize_staged_files(repo, repo_path)
+            if sanitized_files:
+                print(f"   ✓ Sanitized {len(sanitized_files)} staged file(s)")
+            self._verify_staged_files_sanitized(repo)
+
             # Unstage files exceeding GitHub's 100MB file size limit
             large_files = self._unstage_large_files(repo, repo_path)
             if large_files:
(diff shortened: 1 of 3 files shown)


AndrewRqy:
Thank you for working on this. The staged Git boundary is the right place for the final publication check, but this should replace the existing pre-commit log sanitization rather than create a second publication-time path.

NeuriCo already calls `sanitize_logs_directory()` inside `commit_and_push()`. Once the staged-content mechanism is robust, it should become the single authoritative pre-commit check. The real-time output redaction, `.gitignore` rules, credential isolation, and transient Git authentication should remain because they protect different stages.

Before merging, could you please address the following:

1. Replace the existing `sanitize_logs_directory()` call in `commit_and_push()` with the staged-content boundary, while continuing to use the shared patterns in `core.security`.
2. Ensure non-UTF-8 staged files cannot bypass credential verification. Binary artifacts containing recognized credentials should be rejected or unstaged.
3. Avoid following symbolic links during sanitization. Only the staged Git blob should be inspected so files outside the repository cannot be modified.
4. Exclude oversized files before reading staged content into memory, or use bounded or streaming inspection.
5. Rebase onto the latest `main`, preserving the existing transient Git authentication and `push_if_clean` behavior.

This would provide one clear publication boundary while retaining the other safeguards as complementary layers.

mimosapudical:
> Thank you for working on this. The staged Git boundary is the right place for the final publication check, but this should replace the existing pre-commit log sanitization rather than create a second publication-time path.
> 
> NeuriCo already calls `sanitize_logs_directory()` inside `commit_and_push()`. Once the staged-content mechanism is robust, it should become the single authoritative pre-commit check. The real-time output redaction, `.gitignore` rules, credential isolation, and transient Git authentication should remain because they protect different stages.
> 
> Before merging, could you please address the following:
> 
> 1. Replace the existing `sanitize_logs_directory()` call in `commit_and_push()` with the staged-content boundary, while continuing to use the shared patterns in `core.security`.
> 2. Ensure non-UTF-8 staged files cannot bypass credential verification. Binary artifacts containing recognized credentials should be rejected or unstaged.
> 3. Avoid following symbolic links during sanitization. Only the staged Git blob should be inspected so files outside the repository cannot be modified.
> 4. Exclude oversized files before reading staged content into memory, or use bounded or streaming inspection.
> 5. Rebase onto the latest `main`, preserving the existing transient Git authentication and `push_if_clean` behavior.
> 
> This would provide one clear publication boundary while retaining the other safeguards as complementary layers.

Thanks — I addressed all five points in `06d32cb` and rebased the branch directly onto the current `main` (`78940d0`, including #169).

The staged Git index is now the single authoritative pre-commit publication boundary. The old `sanitize_logs_directory()` call has been removed from `commit_and_push()`, while the shared credential patterns remain in `core.security`.

The new implementation operates directly on staged Git blobs rather than working-tree files. Binary/non-UTF-8 blobs are inspected as raw bytes and fail closed when a recognized credential is present; symbolic-link blobs are inspected without following or modifying their filesystem targets; and oversized staged objects are detected and unstaged using Git object size before their content is read.

I also preserved the current `push_if_clean` behavior and transient `http.extraHeader` authentication from `main`.

Regression coverage was added for index-only text sanitization, binary fail-closed behavior, safe binary preservation, symlink safety, oversized-before-read handling, staged deletions, and `push_if_clean` compatibility.

PR #172, opened 2026-08-21
Sanitized scoring-conformance report as manager review evidence

AndrewRqy:
The HITL manager reviews the rule maker's scoring design, but it may not read the sealed evaluator (`scoring/eval.py`, `scoring/targets.json`, `scoring/rule_maker_log.md`). That restriction is intentional: the manager persists into later experiment stages, so exposing the hidden evaluator to it could leak the implementation into worker feedback. The consequence is a gap. Runtime validation catches syntax, ABI, and structural problems, but nothing checks whether the harness actually honors the user's declared evaluation contract. A rule maker that reimplements a required function, ignores its output, or mistranscribes a metric in `targets.json` passes review unnoticed.

This adds an advisory verifier that closes the gap without breaking the restriction. The verifier can read the sealed evaluator; the manager reads only a sanitized report of its conclusions and keeps every decision.

## What this adds

At the rule-maker phase-finish, past the plan phase, the runtime runs the verifier and injects its report into the manager's existing rule-maker review as advisory evidence:

- The manager keeps the accept-or-rerun decision. The verifier never approves, rejects, or blocks the workflow.
- A verifier crash or timeout becomes an `UNAVAILABLE` report, handled as the verifier's own failure, never returned to the rule maker.
- It runs only when the idea declares an evaluation contract, and only for the rule maker.
- On a CONCERNS report the manager sees the specific user-declared requirements that may be unmet, named in the user's own words, so it can steer a rerun precisely.

Like the scorer, it is a runtime-invoked evaluator utility, not a HITL worker, so it needs no durable-worker machinery. This is HITL-only; flat/Standard AutoResearch keeps its existing blocking verifier, which is correct there because that mode has no manager.

## Leak-proofing

The verifier is the only actor allowed to read the sealed evaluator, so isolation is defense in depth:

1. **Scoped sandbox (focus, not a security boundary).** The verifier runs with the run's permissions like every other HITL worker, so per the design doc it is not OS-sandboxed. The throwaway sandbox is a focus mechanism: it holds only the files the verifier needs to judge, the rule maker's outputs under `scoring/` and the staged mandated functions under `code/local/`, copied as regular files with symlinks skipped so the sandbox never itself materializes a pointer to `data/.test/`. Its transcript, logs, and `verification.json` are written inside the sandbox and discarded. Isolation of the result comes instead from the report being leak-proof by construction and from the verifier being advisory: it cannot change or seal anything, so its filesystem access cannot affect scoring.
2. **Report leak-proof by construction.** The manager report is assembled only from a fixed status (PASS / CONCERNS / UNAVAILABLE), canned category descriptions keyed by recognized check names, and, on a CONCERNS, the user's own declared requirements named verbatim (metric names and targets, mandated functions, results format). Those requirements come from `idea.evaluation`, the user's input, which is not sealed. The report never reads the verdict's `summary`, `detail`, or `evidence`, an unrecognized check name, or anything derived from the sealed files, so it gives the manager exactly which declared requirement may be unmet without revealing what the evaluator contains.
3. **Prompt-injection hardening.** The report is derived from a potentially adversarial `eval.py`, so the manager prompt treats it as data, not instructions.

Because the verifier operates entirely on a disposable copy, the conformance path never writes to the reviewed workspace, so there is no partial-state-survives-cleanup class of failure.

The report is persisted as part of the durable phase-finish request and replayed on resume, so a restart continues from the same evidence rather than rerunning the model verifier. The PASS report covers the metrics, targets, required functions, and results format the verifier actually checks, and defers the scientific validity of the evaluation split to the manager.

## Files

| File | Change |
|---|---|
| `src/agents/eval_verifier.py` | `build_manager_conformance_report`: render a verdict as a leak-proof, canned, manager-facing report. |
| `src/core/pipeline_orchestrator.py` | `_scoring_conformance_report`: run the verifier in a scoped sandbox (scoring/ + staged code/local/); wire the reporter onto the rule-maker runtime. |
| `src/core/hitl.py` | Runtime carries the conformance reporter and produces the report for the review, gated past the plan phase, degrading to UNAVAILABLE on error. |
| `src/core/hitl_manager_react.py` | `review_phase_finish` takes the report and passes it to the template. |
| `templates/hitl/manager_review_phase_finish.txt` | Rule-maker review renders the report as advisory evidence, framed as data not instructions. |

## Validation

Live end-to-end run with the real verifier: seeded a workspace with `scoring/{eval.py,targets.json,interface.md}` plus a decoy `data/.test/answers.json` secret, then ran the conformance path. The verifier launched in the sandbox, read only the three copied files, and returned a PASS report in 50s. Post-run: no `verification.json` in the real workspace, the decoy secret untouched, the sandbox removed, and no sealed strings in the report. This also confirmed the verifier works with a sandbox containing only `scoring/`, so it does not need `.venv`, provider skills, or `data/`.

19 focused tests cover the report outcomes, adversarial leak-proofing (hidden values, answer keys, and a prompt-injection payload as a check name, none of which reach the report), sandbox scoping (secret not provided, symlinks skipped, real workspace untouched, sandbox cleaned up on success and on failure), the runtime gating, and the template wiring. Full suite green.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -15,10 +15,6 @@
 # the ordinary non-HITL scoring pipeline treats verification as a gate.
 # OPENROUTER_KEY=
 # OPENAI_API_KEY=
-# Dedicated verifier credentials are never passed to coding agents. If neither
-# is set, HITL reports API NOT AVAILABLE and manager review continues.
-# NEURICO_EVAL_VERIFIER_OPENROUTER_KEY=
-# NEURICO_EVAL_VERIFIER_OPENAI_API_KEY=
 # NEURICO_EVAL_VERIFIER_MODEL=       # Optional verifier model override
 
 # ─── OPTIONAL: Additional API keys (passed through to agent environment) ───
--- a/.env.example
+++ b/.env.example
@@ -9,9 +9,17 @@
 # ─── OPTIONAL: LLM-assisted conversion, naming, paper-finder, and experiments ───
 # OpenRouter is preferred; a direct OpenAI key also works. Without either key,
 # IdeaHub and local text conversion use a template and repository naming uses a
-# basic fallback.
+# basic fallback. A scoring-enabled idea with a declared evaluation contract
+# also uses this key for its tool-less scorer-verification API call. In HITL,
+# a missing/unavailable API is reported to the manager and does not block review;
+# the ordinary non-HITL scoring pipeline treats verification as a gate.
 # OPENROUTER_KEY=
 # OPENAI_API_KEY=
+# Dedicated verifier credentials are never passed to coding agents. If neither
+# is set, HITL reports API NOT AVAILABLE and manager review continues.
+# NEURICO_EVAL_VERIFIER_OPENROUTER_KEY=
+# NEURICO_EVAL_VERIFIER_OPENAI_API_KEY=
+# NEURICO_EVAL_VERIFIER_MODEL=       # Optional verifier model override
 
 # ─── OPTIONAL: Additional API keys (passed through to agent environment) ───
 # Agents may use these during experiments (e.g., for model access, routing)
--- a/README.md
+++ b/README.md
@@ -192,7 +192,8 @@ declared contract and a bounded allowlist of scorer/function source to the
 configured external API; it never
 launches a coding agent. OpenRouter requests require zero-data-retention and
 deny data collection per request. In HITL mode an unavailable verifier is
-reported as `API NOT AVAILABLE` and manager review continues. In non-HITL
+reported as `API NOT AVAILABLE`, a malformed response is reported as
+`VERIFICATION INCONCLUSIVE`, and manager review continues. In non-HITL
 scoring, verification remains a gate and an unavailable API fails the
 rule-maker stage. `COHERE_API_KEY` is optional (improves paper ranking).
 
--- a/README.md
+++ b/README.md
@@ -187,9 +187,9 @@ GitHub; `GITHUB_ORG` is optional (uses the personal account if empty)
 
 **Paper Finder and scoring verifier** — `S2_API_KEY` is required for full
 paper-finder, together with `OPENROUTER_KEY` or `OPENAI_API_KEY`. Scoring-contract
-verification uses a separate `NEURICO_EVAL_VERIFIER_*` credential that is never
-passed to coding agents. The verifier sends the declared contract and a bounded
-allowlist of scorer/function source to the configured external API; it never
+verification uses the same configured OpenRouter or OpenAI API access. The verifier sends the
+declared contract and a bounded allowlist of scorer/function source to the
+configured external API; it never
 launches a coding agent. OpenRouter requests require zero-data-retention and
 deny data collection per request. In HITL mode an unavailable verifier is
 reported as `API NOT AVAILABLE` and manager review continues. In non-HITL
@@ -200,8 +200,6 @@ rule-maker stage. `COHERE_API_KEY` is optional (improves paper ranking).
 | --- | --- | --- |
 | `OPENROUTER_KEY` | Yes, unless `OPENAI_API_KEY` is set | OpenRouter access for paper-finder, IdeaHub conversion, LLM repo naming, and experiments that need it |
 | `OPENAI_API_KEY` | Yes, unless `OPENROUTER_KEY` is set | Direct OpenAI access for paper-finder, IdeaHub conversion, LLM repo naming, and experiments that need it |
-| `NEURICO_EVAL_VERIFIER_OPENROUTER_KEY` | For verifier, unless its OpenAI key is set | Dedicated OpenRouter verifier credential; never passed to coding agents |
-| `NEURICO_EVAL_VERIFIER_OPENAI_API_KEY` | For verifier, unless its OpenRouter key is set | Dedicated direct-OpenAI verifier credential; never passed to coding agents |
 | `NEURICO_EVAL_VERIFIER_MODEL` | No | Override the verifier model (`openai/gpt-4.1` through OpenRouter or `gpt-4.1` through OpenAI by default) |
 | `S2_API_KEY` | For paper-finder | Semantic Scholar API key ([get here](https://www.semanticscholar.org/product/api)) |
 | `COHERE_API_KEY` | No | Improves paper-finder ranking (~7% boost) |
--- a/README.md
+++ b/README.md
@@ -183,16 +183,26 @@ GitHub; `GITHUB_ORG` is optional (uses the personal account if empty)
 | `GITHUB_ORG` | No | GitHub org name (default: personal account) |
 
 <details>
-<summary>Paper Finder and agent API keys</summary>
+<summary>Paper Finder, scoring verifier, and agent API keys</summary>
 
-**Paper Finder** — `S2_API_KEY` and either `OPENROUTER_KEY` or
-`OPENAI_API_KEY` required for full paper-finder; `COHERE_API_KEY` optional
-(improves ranking)
+**Paper Finder and scoring verifier** — `S2_API_KEY` is required for full
+paper-finder, together with `OPENROUTER_KEY` or `OPENAI_API_KEY`. Scoring-contract
+verification uses a separate `NEURICO_EVAL_VERIFIER_*` credential that is never
+passed to coding agents. The verifier sends the declared contract and a bounded
+allowlist of scorer/function source to the configured external API; it never
+launches a coding agent. OpenRouter requests require zero-data-retention and
+deny data collection per request. In HITL mode an unavailable verifier is
+reported as `API NOT AVAILABLE` and manager review continues. In non-HITL
+scoring, verification remains a gate and an unavailable API fails the
+rule-maker stage. `COHERE_API_KEY` is optional (improves paper ranking).
 
 | Variable | Required | Description |
 | --- | --- | --- |
-| `OPENROUTER_KEY` | Yes, unless `OPENAI_API_KEY` is set | OpenRouter access for paper-finder, IdeaHub conversion, and LLM repo naming |
-| `OPENAI_API_KEY` | Yes, unless `OPENROUTER_KEY` is set | Direct OpenAI access for paper-finder, IdeaHub conversion, and LLM repo naming |
+| `OPENROUTER_KEY` | Yes, unless `OPENAI_API_KEY` is set | OpenRouter access for paper-finder, IdeaHub conversion, LLM repo naming, and experiments that need it |
+| `OPENAI_API_KEY` | Yes, unless `OPENROUTER_KEY` is set | Direct OpenAI access for paper-finder, IdeaHub conversion, LLM repo naming, and experiments that need it |
+| `NEURICO_EVAL_VERIFIER_OPENROUTER_KEY` | For verifier, unless its OpenAI key is set | Dedicated OpenRouter verifier credential; never passed to coding agents |
+| `NEURICO_EVAL_VERIFIER_OPENAI_API_KEY` | For verifier, unless its OpenRouter key is set | Dedicated direct-OpenAI verifier credential; never passed to coding agents |
+| `NEURICO_EVAL_VERIFIER_MODEL` | No | Override the verifier model (`openai/gpt-4.1` through OpenRouter or `gpt-4.1` through OpenAI by default) |
 | `S2_API_KEY` | For paper-finder | Semantic Scholar API key ([get here](https://www.semanticscholar.org/product/api)) |
 | `COHERE_API_KEY` | No | Improves paper-finder ranking (~7% boost) |
 
--- a/README.md
+++ b/README.md
@@ -195,7 +195,9 @@ deny data collection per request. In HITL mode an unavailable verifier is
 reported as `API NOT AVAILABLE`, a malformed response is reported as
 `VERIFICATION INCONCLUSIVE`, and manager review continues. In non-HITL
 scoring, verification remains a gate and an unavailable API fails the
-rule-maker stage. `COHERE_API_KEY` is optional (improves paper ranking).
+rule-maker stage. The verifier performs one request with SDK retries disabled
+and enforces an outer wall-clock deadline over the complete request.
+`COHERE_API_KEY` is optional (improves paper ranking).
 
 | Variable | Required | Description |
 | --- | --- | --- |
--- a/docs/HITL_AUTORESEARCH.md
+++ b/docs/HITL_AUTORESEARCH.md
@@ -389,8 +389,9 @@ request declares no shell, filesystem, network, MCP, function, or write tools.
 The remote response is strictly parsed by runtime and reduced immediately to
 fixed check categories. HITL writes no prompt, raw response, verifier verdict,
 or verifier audit file into the workspace; the
-durable manager handoff contains only `PASS`, `CONCERNS`, or
-`API NOT AVAILABLE`, fixed runtime-owned descriptions, and names from the
+durable manager handoff contains only `PASS`, `CONCERNS`,
+`VERIFICATION INCONCLUSIVE`, or `API NOT AVAILABLE`, fixed runtime-owned
+descriptions, and names from the
 user's own declarations. Model-generated detail is not passed to the manager
 or to a later coding-agent prompt.
 
@@ -404,10 +405,11 @@ is acceptable for that source.
 The report is advisory in HITL mode. If the API key is absent or the provider is
 unreachable, the manager sees `API NOT AVAILABLE` and continues the normal
 public-design review. That status is never represented as a pass. A provider
-response that arrives but cannot be validated produces `CONCERNS`, since it did
-not establish conformance. Local evidence failures are also different: missing,
-unsafe, non-UTF-8, or oversized rule-maker artifacts produce a fixed `CONCERNS`
-report so a worker cannot manufacture an apparent provider outage. Final rule-
+response that arrives but cannot be validated produces `VERIFICATION
+INCONCLUSIVE`: it did not establish conformance, but it is not evidence of a
+scoring-design defect. Local evidence failures are different: missing, unsafe,
+non-UTF-8, or oversized rule-maker artifacts produce a fixed `CONCERNS` report
+so a worker cannot manufacture an apparent provider outage. Final rule-
 maker approval also rechecks the public workspace fingerprint captured for the
 review, preventing changed artifacts from being sealed under a stale report.
 
--- a/docs/HITL_AUTORESEARCH.md
+++ b/docs/HITL_AUTORESEARCH.md
@@ -371,22 +371,24 @@ meaning of the result.
 When the idea declares metrics, a results format, or a required evaluation
 function, manager review of the rule-maker checkpoint receives an automated
 conformance report. Trusted runtime code reads only `scoring/eval.py`,
-`scoring/targets.json`, `scoring/interface.md`, `scoring/rule_maker_log.md`, and
-the specifically declared staged function files. It rejects symlinks, path
+`scoring/targets.json`, `scoring/interface.md`, optional
+`scoring/rule_maker_log.md` when present, and the specifically declared staged
+function files. It rejects symlinks, path
 traversal, non-UTF-8 input, and evidence above the per-file or total byte caps,
 then submits that bundle and the user's declared evaluation contract to the
 configured OpenRouter or OpenAI API.
 
-Verifier access uses a dedicated API credential that runtime removes from every
-coding-agent environment. OpenRouter calls additionally require a zero-data-
-retention endpoint and deny provider data collection on each request. If those
+Verifier access uses the configured OpenRouter or OpenAI API credential.
+OpenRouter calls additionally require a zero-data-retention endpoint and deny
+provider data collection on each request. If those
 constraints cannot be met, the call fails as `API NOT AVAILABLE` rather than
 routing the scorer source through a less restrictive endpoint.
 
 This verifier is a tool-less model request, not a coding-agent process. The
 request declares no shell, filesystem, network, MCP, function, or write tools.
-The remote response is strictly parsed by runtime. HITL writes no prompt, raw
-response, verifier verdict, or verifier audit file into the workspace; the
+The remote response is strictly parsed by runtime and reduced immediately to
+fixed check categories. HITL writes no prompt, raw response, verifier verdict,
+or verifier audit file into the workspace; the
 durable manager handoff contains only `PASS`, `CONCERNS`, or
 `API NOT AVAILABLE`, fixed runtime-owned descriptions, and names from the
 user's own declarations. Model-generated detail is not passed to the manager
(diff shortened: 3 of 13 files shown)


Frankbest18 (changes requested):
Thanks Andrew. I agree with the main direction here: the verifier is advisory, the manager keeps the decision, and a verifier failure becomes `UNAVAILABLE` rather than blocking or rolling back the workflow. That is much closer to the authority model we use throughout HITL.

I still found several issues that I think need to be addressed before this is ready:

**[P1] The verifier is not actually isolated from the real workspace**

The temporary directory limits which files are placed in the verifier’s working directory, but the verifier is still launched with the run’s `full_permissions` setting. In a normal full-permission run, that means `--yolo` or `--dangerously-skip-permissions`. The process can still read or modify files outside the temporary directory, including the original workspace and `data/.test/`.

This matters because the verifier is a model agent reading potentially adversarial file content. Changing its working directory is not a runtime-enforced security boundary. The current tests only show that the secret was not copied into the temporary directory; they do not show that the verifier process cannot access the secret at its original path.

**[P2] The routing check does not receive the files it needs**

The verifier’s prompt says it must inspect the staged mandated functions under `code/local/` to determine whether `eval.py` genuinely calls them and uses their results. The temporary workspace, however, contains only the four rule-maker outputs under `scoring/`.

For any contract with a `required_for_evaluation` function, the verifier therefore cannot perform the routing check as specified. Any PASS or CONCERNS result for that check would be based on incomplete evidence.

**[P2] The verifier report is not part of the durable manager request**

The report is generated immediately before manager review and inserted into the prompt, but it is not stored in the pending worker command or included in the request identity.

If NeuriCo restarts while the manager request is pending, recovery can rerun the model verifier and produce a different report for the same workspace and request key. The manager may then receive conflicting evidence for what HITL considers one durable review request. It can also repeat an expensive verifier call during recovery.

This is inconsistent with the existing HITL restart model, where a resumed request should continue from the same persisted facts rather than regenerate nondeterministic evidence.

**[P2] The PASS report claims that the evaluation split was checked**

The manager-facing PASS says the verifier confirmed the metrics, targets, evaluation split, and required functions. The verifier currently defines only routing, transcription, and format checks; it does not check the evaluation split.

That wording can give the manager confidence in a conclusion the verifier never established, particularly since the manager is expected to review the scientific validity of the documented split itself.

The existing `pending_worker_command` and `resume_pending_worker_command` paths are useful references for the evidence-lifecycle issue, and the current HITL scoring workspace and workspace-guard code show the kind of mechanically enforced boundary we normally rely on for isolation.

AndrewRqy:
Thanks for this detailed review, i will try to address the four in the following.

[P1] You're right that the working directory is not a runtime-enforced boundary, and after some thinking i dont think gating the agent from accessing the files is as effective as gating it from producing any modification beside the one report it writres for manager. This means that instead of preventing the verifier from seeing the sealed files, we prevent it from doing anything besides writing the report, which is in fixed format and should not disclose any information the manager has no access to. The sandbox is a focus mechanism: it gives the verifier only the files it needs to judge and discards its writes, but the verifier runs with the run's permissions like every other HITL worker, which per the doc is not an OS sandbox. The guarantee that actually matters is not that it cannot read a given file but that it edits nothing besides its verdict, so I leaned on the read-only enforcement run_eval_verifier already has: it snapshots the reviewed files and, if the agent edits or creates any, restores them and fails the verdict as read_only. It only covered the three scoring/ files, so I extended it to the staged mandated functions under code/local/ too. A verifier that edits anything it was given is restored and fails read-only, and a read_only verdict now reports UNAVAILABLE since a verifier that tampered cannot be trusted to have judged. I also softened the comments and tests so they no longer claim the secret is unreachable, only that it is not provided in the sandbox.

[P2 routing] The staged mandated functions under code/local/ are now copied into the sandbox alongside scoring/, as regular files with symlinks skipped, so the routing check has the evidence its prompt requires.

[P2 durable report] The report is now persisted as part of the durable phase-finish request and replayed on resume, so a restart continues from the same evidence instead of rerunning the model verifier. It is stored on the pending command but kept out of the request-key identity, so the key stays deterministic.

[P2 PASS wording] The PASS report no longer claims the evaluation split was checked. It covers the metrics, targets, required functions, and results format the verifier actually checks, and defers the scientific validity of the split to your review.

Tests cover the code/local/ staging and symlink-skip, the durable replay (a resumed request does not rerun the verifier), the read-only guarantee (a verifier that edits a staged file is restored and fails read-only), and the corrected PASS wording.

Frankbest18 (changes requested):
Thanks for the update, Andrew. I think the main P1 is still there. Calling the temporary directory a focus mechanism is fair, but it does not make the write-safety issue go away.

**[P1] The verifier runs after the rule maker’s workspace has already been validated**

Runtime fingerprints the workspace and validates the rule-maker output first. It then starts the verifier with full permissions. If the verifier changes the real workspace, that change happens after validation, and the manager may approve a workspace different from the one runtime actually checked.

The verifier being advisory only means it cannot make the review decision. It does not prevent it from changing the files involved in that decision.

**[P1] The restore code can write through a verifier-created symlink**

The new read-only check uses `is_file()` and then restores the original content with `write_bytes()`. Both follow symlinks. If the verifier replaces a reviewed file with a symlink, the restore writes to the symlink target, which may be a file in the real workspace.

In that case, NeuriCo’s own recovery code ends up overwriting a path chosen by the verifier.

**[P2] The read-only check still does not cover everything the verifier can change**

The verifier reads `interface.md` for its format check, but `interface.md` is not included in `reviewed_files`. It can change that file, review the changed version, and return PASS without the guard noticing. New files created under `code/local/` are also not covered because the guard records only the files that already exist.

My larger concern is that this creates a second protection and recovery path inside HITL. We already have established mechanisms for this:

- `HitlStageRollback` manages the lifecycle and rollback of model-driven stages.
- `HitlWorkspaceWriteGuard` checks the relevant workspace boundary, including symlinks and unexpected files.
- `HitlGitStateStore` provides durable recovery.
- `hitl_scoring_workspace.py` provides runtime-owned isolation for scoring.

The temporary workspace and the new per-file snapshot/restore logic overlap with these mechanisms, but they do not provide the same guarantees. The symlink overwrite and incomplete file coverage above are examples of the problems this inconsistency introduces.

I think we need to be careful here. The verifier can remain advisory evidence for the manager, but its execution should follow the same lifecycle and state-integrity model as the rest of HITL. We should not add a parallel mechanism for protecting and recovering workspace state. This workflow is already complex, and having two different ways to handle the same problem will make it harder to reason about and risks weakening the reliability we have already built.

AndrewRqy:
Adopted the workspace guard HITL already has instead of the per-file snapshot, which was operating on the discarded sandbox copy and so never protected the real workspace.

[P1] _scoring_conformance_report now wraps the verifier run in HitlWorkspaceWriteGuard: capture_public before, require_unchanged after. If the verifier changes the real reviewed workspace, that is detected and the report is UNAVAILABLE, so the manager never approves a workspace different from the one runtime validated.

[P1 symlink] Reverted the per-file guard extension, so we stop growing a second protect-and-recover path. The remaining narrow guard (flat-mode safeguard over the sealed scoring files) now unlinks a symlink before restoring, so it never writes through a verifier-planted link into another path.

[P2 coverage] With the workspace guard doing the check, interface.md and any newly-created files are covered, since it fingerprints the whole public boundary rather than an enumerated file list.

The sandbox stays purely as a focus mechanism, and write safety comes from the same guard the rest of HITL uses. Tests cover a verifier that mutates the real workspace being reported UNAVAILABLE, a clean run passing, and the restore not writing through a symlink. Pushed as 6058854.

Frankbest18 (changes requested):
Thanks for the update. Reusing `HitlWorkspaceWriteGuard` is closer to the existing HITL design, but I still see two P1 issues.

**[P1] Verifier mutations remain active after they are detected**

`require_unchanged()` detects a change, but it does not restore the workspace. The current path changes the verifier report to `UNAVAILABLE` and continues to manager review. The manager is told that an unavailable verifier should not prevent an independent approval, so the stage can still complete and discard its rollback snapshot.

The result is that the manager may approve a workspace different from the one runtime validated. Marking the verifier’s report unavailable only invalidates its judgment; it does not address the state change that caused the report to become unavailable.

**[P1] The guard does not cover the verifier’s full write authority**

`capture_public()` deliberately excludes `.neurico`, `.git`, `.venv`, the provider directories, logs, and other runtime-owned state. That is appropriate for its normal role: checking a worker’s public write boundary without treating NeuriCo’s own runtime writes as violations.

The verifier runs with full permissions, however, and the temporary directory is explicitly not a security boundary. It can modify any of those excluded locations without the guard noticing. A change to `.neurico`, `.git`, or `.venv` could affect workflow recovery, repository state, or the environment used by scoring while the verifier report still returns `PASS`.

Calling the report “advisory” does not make the verifier process itself harmless. “Advisory” only describes how the manager should interpret the report. For that model to be safe, producing the report must not be able to change the validated workspace or NeuriCo’s runtime state. At the moment, the report is advisory, but the model process producing it still has authority over the state being reviewed.

The most relevant references for this path are:

- [`[docs/HITL_AUTORESEARCH.md](https://github.com/ChicagoHAI/NeuriCo/blob/60588541f15828fef3f17d9f0b9d3fcbade13b81/docs/HITL_AUTORESEARCH.md#L35-L127)`](https://github.com/ChicagoHAI/NeuriCo/blob/60588541f15828fef3f17d9f0b9d3fcbade13b81/docs/HITL_AUTORESEARCH.md#L35-L127), especially the workflow invariants, actor authority, and stage lifecycle.
- [`[src/core/hitl.py](https://github.com/ChicagoHAI/NeuriCo/blob/60588541f15828fef3f17d9f0b9d3fcbade13b81/src/core/hitl.py#L2832-L2875)`](https://github.com/ChicagoHAI/NeuriCo/blob/60588541f15828fef3f17d9f0b9d3fcbade13b81/src/core/hitl.py#L2832-L2875), where the workspace is fingerprinted and validated before manager review.
- [`[src/core/hitl_stage_runtime.py](https://github.com/ChicagoHAI/NeuriCo/blob/60588541f15828fef3f17d9f0b9d3fcbade13b81/src/core/hitl_stage_runtime.py#L22-L165)`](https://github.com/ChicagoHAI/NeuriCo/blob/60588541f15828fef3f17d9f0b9d3fcbade13b81/src/core/hitl_stage_runtime.py#L22-L165), which defines the existing worker-recovery and stage-rollback lifecycle.
- [`[src/core/hitl_scoring_workspace.py](https://github.com/ChicagoHAI/NeuriCo/blob/60588541f15828fef3f17d9f0b9d3fcbade13b81/src/core/hitl_scoring_workspace.py#L70-L177)`](https://github.com/ChicagoHAI/NeuriCo/blob/60588541f15828fef3f17d9f0b9d3fcbade13b81/src/core/hitl_scoring_workspace.py#L70-L177), which shows how NeuriCo evaluates an approved checkpoint without exposing the live workspace to the evaluator.
- [`[src/core/hitl_workspace_guard.py](https://github.com/ChicagoHAI/NeuriCo/blob/60588541f15828fef3f17d9f0b9d3fcbade13b81/src/core/hitl_workspace_guard.py#L14-L101)`](https://github.com/ChicagoHAI/NeuriCo/blob/60588541f15828fef3f17d9f0b9d3fcbade13b81/src/core/hitl_workspace_guard.py#L14-L101), which documents the public-only scope and detect-only behavior of the guard.

This path still allows an unmanaged model process to alter state after validation and before approval. That breaks the stable review boundary HITL is designed to preserve, so I do not think it is safe to merge yet.

AndrewRqy:
Thanks for the review:) Reworked P1 to prevent the writes by capability rather than detect them after, which I think fully addresses both parts.

[P1 write authority] The verifier no longer runs with --dangerously-skip-permissions. It runs confined by the provider CLI so it cannot write outside its throwaway sandbox, and therefore cannot change the reviewed workspace or runtime state (.neurico, .git, .venv). For claude, a dontAsk permission profile with no Bash and Write/Edit scoped to the one verdict file, so it can only write scoring/verification.json. For codex, its OS-level workspace-write sandbox, which confines writes to the working sandbox and system temp, both outside the reviewed workspace. The restriction is provider-enforced, so the report is produced only for providers that support it; for any other provider we offer no report rather than run an unconfined agent. I verified both against the real CLIs: the verdict write is allowed, writes to a sibling file, /tmp, and a home-dir path are blocked and non-blocking, and an end-to-end claude run produced a report with zero denials and the workspace untouched.

[P1 detect vs restore] Since the mutation is now prevented, there is nothing to detect-and-restore. I kept HitlWorkspaceWriteGuard only as defense in depth, reframed as an invariant tripwire behind the restriction, not the protection itself: it fingerprints the reviewed workspace around the run, and if the confinement ever regressed and the verifier changed the workspace, it logs loudly and reports UNAVAILABLE so an escaped run is never trusted. The per-file guard extension is removed.

Frankbest18 (changes requested):
Thanks, Andrew. Preventing verifier writes is a much better direction than detecting and repairing them afterward. I still see three P1 issues in the current implementation.

**[P1] Codex is not confined to the verifier directory**

`--sandbox workspace-write` also permits system temporary directories by default and can inherit additional writable roots from user configuration. The verifier sandbox and NeuriCo’s isolated scoring workspaces are both created in the system temporary area. The verifier can therefore write outside its own directory, leave sealed material behind after cleanup, or affect another temporary NeuriCo workspace.

The current tests only check that `--sandbox workspace-write` appears in the command. They do not verify the effective writable roots. These settings are documented in the [[Codex workspace-write configuration](https://github.com/openai/codex/blob/main/codex-rs/app-server-protocol/schema/json/v2/ConfigReadResponse.json)](https://github.com/openai/codex/blob/main/codex-rs/app-server-protocol/schema/json/v2/ConfigReadResponse.json).

**[P1] The Claude permission profile can be widened by loaded configuration**

`--allowedTools` adds rules alongside permissions loaded from settings; it does not replace them. Claude also runs hooks before normal permission evaluation, and a hook can approve a tool call. Since this invocation still loads those settings and hooks, an existing broader `Write` or `Edit` permission can expand the verifier’s authority beyond `verification.json`.

The tests confirm that the intended flags appear in the command, but they do not verify the effective permission set after configuration and hooks are loaded. This behavior is described in the [[Claude CLI reference](https://docs.anthropic.com/en/docs/claude-code/cli-usage)](https://docs.anthropic.com/en/docs/claude-code/cli-usage) and [[Claude permissions documentation](https://docs.anthropic.com/en/docs/claude-code/iam)](https://docs.anthropic.com/en/docs/claude-code/iam).

**[P1] A confinement failure still leaves the validated state unprotected**

The workspace guard remains a detect-only check over the public workspace. If confinement fails and the verifier changes a public file, the report becomes `UNAVAILABLE`, but the change remains active. The manager can continue, approve the stage, and discard the rollback snapshot. Changes to excluded runtime state are not detected at all.

In that situation, it is not enough to distrust the verifier report; the workspace itself may no longer be the one runtime validated.

More broadly, this is creating a separate integrity model for the verifier. HITL already defines how this boundary should work: runtime owns validation and recovery, the manager reviews a fixed validated state, and sensitive evaluation runs from an approved checkpoint in a runtime-owned isolated workspace. Those contracts are described in the [[HITL invariants and stage lifecycle](https://github.com/ChicagoHAI/NeuriCo/blob/9c751100fbb61f6bf3526722d34fbba1a1461de4/docs/HITL_AUTORESEARCH.md#L35-L127)](https://github.com/ChicagoHAI/NeuriCo/blob/9c751100fbb61f6bf3526722d34fbba1a1461de4/docs/HITL_AUTORESEARCH.md#L35-L127), the [[validation-to-manager-review handoff](https://github.com/ChicagoHAI/NeuriCo/blob/9c751100fbb61f6bf3526722d34fbba1a1461de4/src/core/hitl.py#L2832-L2875)](https://github.com/ChicagoHAI/NeuriCo/blob/9c751100fbb61f6bf3526722d34fbba1a1461de4/src/core/hitl.py#L2832-L2875), and the existing [[isolated scoring workspace](https://github.com/ChicagoHAI/NeuriCo/blob/9c751100fbb61f6bf3526722d34fbba1a1461de4/src/core/hitl_scoring_workspace.py#L70-L177)](https://github.com/ChicagoHAI/NeuriCo/blob/9c751100fbb61f6bf3526722d34fbba1a1461de4/src/core/hitl_scoring_workspace.py#L70-L177).

Rather than adding more provider-specific confinement, I’d bring the verifier into that same lifecycle and integrity model. Otherwise we end up with one set of guarantees for the rest of HITL and another, provider-dependent set for this verifier.

AndrewRqy:
Thanks for the review. I replaced the coding-agent verifier with a tool-less API call.

The verifier now:

- Receives only an allowlisted and size-limited evidence bundle.
- Has no filesystem, shell, MCP, function-calling, or write capabilities.
- Uses credentials that are not passed to coding agents.
- Persists no raw prompt, response, transcript, audit record, or verdict in HITL.
- Passes only fixed runtime-generated categories to the manager.
- Returns `API NOT AVAILABLE` when the API or a compliant provider is unavailable. Manager review then continues normally.

I also want to point out an inconsistency in the adversarial assumptions applied here. The verifier was treated as a potentially hostile same-user process because full permissions could let it modify `.neurico`, `.git`, or `.venv`. Ordinary HITL workers run with the same full-permission provider modes, while `HitlWorkspaceWriteGuard.capture_public()` explicitly excludes those locations.

I verified this with a temporary local test. A simulated worker corrupted `.neurico/hitl/runtime.json`. The workspace guard still reported the boundary as unchanged, and HITL subsequently failed to load the runtime state. The test was only used to verify the behavior and was not added to this PR.

This means the hostile-process concern raised for the verifier also exists in the current ordinary HITL worker model. The API redesign removes that concern specifically for the verifier because there is no longer a local verifier process. But i do think we need to set a standard for the level of adversarial behavior we guard against.

Frankbest18 (changes requested):
Thanks for reworking this, Andrew. I agree with removing the full-permission verifier agent. After reading the new version, though, I wonder if we are creating a separate solution for something we already solved in the manager path.

The interactive backend used by the manager is already read-only: Codex runs in its read-only sandbox, and Claude runs in safe mode with native tools disabled. I think the verifier can use the same backend in a new one-shot session with no tools. Runtime can pass in the evidence built by `generate_eval_verifier_messages()` and sanitize the response as it does now. The verifier should not share the manager’s conversation or transcript, since the manager must never receive the sealed evaluator content.

`LLMBackend`, `HitlManager._backend_for_provider()`, and `HitlManager._send_once()` should provide most of what is needed here. Reusing that path would also avoid adding another set of provider credentials and environment rules.

The current separate API path has already introduced a serious problem:

**[P1] The verifier credentials are not limited to the verifier**

The keys are removed only by `build_agent_environment()`. The scorer passes the full environment to `scoring/eval.py`, which is written by the rule maker, and the paper writer also inherits the full environment. Both can therefore receive and use the verifier credentials.

Adding more filtering to individual callers would be easy to miss again. This is one reason I think the existing read-only backend is the better place for the verifier.

I also noticed two correctness issues:

**[P1] `rule_maker_log.md` is optional during validation but mandatory during verification**

The existing rule-maker validator accepts an output without `scoring/rule_maker_log.md`, while the verifier always tries to read it. The same output can pass validation and then fail before the verifier is called. HITL turns that into `API NOT AVAILABLE`, while flat AutoResearch fails the rule-maker stage.

**[P2] The manager can be shown the wrong failed category**

The verdict parser does not check that the listed violations match the checks marked as failed. For example, the response can mark `routing` as failed but provide only a `transcription` violation. Runtime accepts it, and the manager sees only the transcription concern. Since the manager cannot inspect the sealed evaluator, the actual routing failure is then lost.

AndrewRqy:
Thanks. I considered reusing LLMBackend, but its manager configuration does not provide the same boundary as the direct API path.
For Codex, LLMBackend launches a local codex exec process in a read-only sandbox. Read-only prevents workspace writes, but the process still inherits the local working directory, environment, user configuration, and readable filesystem access. For Claude, safe mode disables native tools and local hooks, but the verifier would still depend on a locally configured CLI process correctly enforcing that restriction.
This matters because the verifier receives sealed evaluator source. A one-shot manager backend would avoid sharing the manager conversation, but it would still introduce a local provider process into the verifier boundary. It may also create provider session or diagnostic state, and HitlManager._send_once() currently has no wall-clock timeout.
The direct API path has a smaller capability boundary. Trusted runtime code selects and bounds the evidence, then sends only that bundle to the provider. The model has no local process, filesystem, shell, MCP, function calls, or manager session. Runtime requests structured JSON, strictly validates the response, and immediately reduces it to fixed categories. HITL persists neither the raw request nor the raw response.
The credential concern no longer requires a separate environment policy. I removed the verifier-specific keys. The verifier now uses the existing configured OpenRouter or OpenAI key, so there is no additional privileged credential for the scorer or paper writer to inherit. OpenRouter requests require ZDR and deny provider data collection. Direct OpenAI requests use store=False.
I also fixed the correctness issues:
- rule_maker_log.md is optional during both validation and verification.
- Violations must match checks that are failed and applicable.
- Missing failed categories are reconstructed from the checks map.
- Malformed provider responses produce CONCERNS, not API NOT AVAILABLE.
- Local evidence failures receive one rule-maker repair attempt in non-HITL mode.
- Actual API unavailability remains advisory in HITL mode.

Frankbest18 (changes requested):
Thanks, Andrew. Your explanation of the direct API boundary makes sense. Therefore, I can notice two smaller issues remaining.

**[P2] A malformed verdict can be presented as a concern about the scoring design**

The parser accepts any JSON object, so a response such as `{}` gets past the initial parsing step despite missing all the required verdict fields. The later validation catches the malformed structure, but `run_eval_verifier()` still returns `success=True` and `passed=False`.

The manager then receives a generic report saying that a declared evaluation requirement may not be met. That is different from what actually happened: the verifier failed to produce a valid review. The report remains advisory in both AutoResearch modes, but it does not give the manager an accurate account of where the concern came from.

**[P2] The verifier timeout does not cover the complete API operation**

The OpenAI client uses two automatic retries by default, and this path does not override that policy. A single verifier invocation can therefore make up to three requests, with the configured timeout applying separately to each attempt.

The operation can remain blocked substantially longer than the reported 180 seconds and may repeat an expensive request containing the full evidence bundle. This does not create an unbounded loop, but the effective behavior does not match the stated one-call, 180-second boundary.

AndrewRqy:
Thanks for catching these! Both issues are addressed in 35e6670.
- Malformed or internally inconsistent verifier responses now produce VERIFICATION INCONCLUSIVE, rather than being presented as a concern about the scoring design. Genuine routing, transcription, or format failures still produce CONCERNS, while API failures remain API NOT AVAILABLE.
- OpenAI and OpenRouter clients now use max_retries=0, so each verifier invocation makes only one request and the configured timeout is no longer applied separately across hidden SDK retries.
I also added regression coverage for empty, invalid-JSON, and structurally malformed verdicts, plus both API client paths. The full root suite passes with 225 tests.

Frankbest18 (changes requested):
Thanks, Andrew. I think only one tiny concern remains:

**[P2] The configured timeout is still not an end-to-end deadline**

The OpenAI client passes `timeout=180` to HTTPX. HTTPX treats this as separate connect, write, read, and pool timeouts. In particular, the read timeout limits how long it may wait between response chunks; it does not limit the total duration of the response.

This means the verifier now makes only one request, but that request can still run for longer than the reported 180 seconds as long as network activity continues. There is no outer wall-clock deadline around the API call. The distinction is described in the [[HTTPX timeout documentation](https://www.python-httpx.org/advanced/timeouts/)](https://www.python-httpx.org/advanced/timeouts/).

AndrewRqy:
Thanks for catching this! 

The verifier now uses the async OpenAI client with an outer wall-clock deadline around the complete API operation. When the deadline expires, runtime cancels the in-flight HTTP request and closes the client. The existing HTTPX phase timeouts and disabled SDK retries remain as secondary safeguards.

I also added regression coverage confirming that a request exceeding the deadline is cancelled and its client is closed. The full root suite passes with 226 tests.

Frankbest18 (changes requested):
Thanks for the update. I did one more overall pass, and there is one recovery issue I think we should address before merging.

**[P2] The persisted verifier report is not replayed after a restart**

`HitlRuntime` and `HitlManager` generate different request keys for the same phase-finish request. Because those keys do not match, recovery does not recognize the report already stored on the pending manager request and calls the verifier again.

This only affects a restart during manager review, but it can cause an unnecessary external model call and give the resumed review a different advisory report from the one in durable state. The current test misses this because it uses the same literal key on both sides rather than the keys produced by the runtime and manager.

AndrewRqy:
Thanks for catching this! The runtime and manager were independently hashing similar but non-identical phase-finish payloads, so their request keys could never match during recovery.
Runtime now owns the phase-finish identity and passes its exact request key to the manager for durable persistence. On restart, the pending command therefore matches the recomputed runtime request and the stored verifier report is replayed without another external call.

Frankbest18 (approved):

PR #173, opened 2026-09-01
fix(hitl): strengthen scoring recovery and evaluation boundaries

Frankbest18:
## Summary

This PR makes initial scoring failures recoverable by returning each defect to the worker that owns it. Experiment defects continue through the experiment-runner repair path. Evaluator and scoring-contract defects restore the pre-experiment workspace and sealed scoring files, then return the concrete failure evidence to the rule maker for correction.

It also defines the evaluation responsibilities of the rule maker, experiment runner, and manager; reproduces nested repository contents inside isolated scoring workspaces; waits for required manager MCP tools before processing begins; and removes fixed execution limits from Full and Auto HITL.

## Initial scoring repair

The initial scoring review now requires the manager to select a `repair_target`.

`repair_target="experiment_runner"` keeps the existing experiment-repair behavior.

`repair_target="rule_maker"` restores the pre-experiment checkpoint, restores the rule maker’s sealed evaluator files, and reopens rule-maker execution with the scoring failure as revision feedback. Once the repair is accepted, NeuriCo reruns the experiment and initial scoring.

A low score alone is not treated as an evaluator defect. The rule maker is selected only when the scoring evidence identifies a problem in the evaluator or scoring-contract implementation.

When the manager sends a failure back to the rule maker, its feedback must describe the concrete conditions that produced the failure and require a matching preflight regression. The feedback is self-contained so the rule maker can reproduce the failure after the workspace has been restored.

## Rule-maker responsibilities

The rule maker is now explicitly responsible for the complete evaluator and every dependency, artifact, and input it requires. Work required by `scoring/eval.py` cannot be deferred to the experiment runner, runtime, or another component.

The rule maker may create, generate, replace, and write required files under `data/.test`, but it may not read their contents. Running `scoring/eval.py` as a functional preflight is permitted and required.

Before completing the stage, the rule maker must verify that the evaluator runs successfully. It must also repeat the preflight in a clean copy containing the same scoring-workspace files but no repository metadata. This prevents an evaluator from depending on nested `.git` state or other information that will not exist during isolated scoring.

Runtime validation now returns complete correction instructions when the rule maker submits incomplete work. Missing inputs, invalid artifacts, interface violations, and preflight failures are reported together. The rule maker is instructed to correct every reported problem, rerun the preflight, and submit phase completion again.

The manager receives the same responsibility boundaries during rule-maker plan review, completion review, and decision resolution. It must reject work that defers evaluator preparation, relies on repository metadata, or lacks functional preflight evidence.

## Scoring interface and experiment ownership

`scoring/interface.md` is limited to the public contract the experiment runner needs to produce its artifacts. It may describe required artifact paths, public entry points, and expected behavior. It must not contain evaluator invocation instructions, evaluator-owned paths, private-input handling, result handling, sealing, or runtime scoring operations.

Within `scoring/`, the experiment runner is instructed to use only `scoring/interface.md`. It must not inspect, depend on, or raise concerns about other scoring files.

Official evaluation is outside the experiment runner’s responsibility and authority. Evaluator availability and official results cannot become experiment prerequisites, checkpoints, blockers, validation steps, or completion conditions.

The manager enforces this boundary during both experiment review and decision resolution. If a worker-authored plan or request assigns official evaluation to the experiment runner, the manager must treat that premise as invalid and require it to be corrected. The same boundary applies during initial experimentation and AutoResearch iterations.

## Isolated scoring workspaces

A checkpoint can contain nested repositories represented as Gitlinks. Creating a normal Git worktree restores the Gitlink entry but not the nested repository contents that were present when the rule maker ran its preflight.

The isolated scorer now enumerates Gitlinks from the selected checkpoint with `git ls-tree`. It validates each Gitlink path and commit against the corresponding nested repository in the source workspace, archives that exact commit, and safely extracts its files into the isolated scoring workspace.

The scorer therefore receives the same dependency contents used during rule-maker preflight without receiving nested repository metadata.

Temporary scoring worktrees are now created beside the configured workspace root. Docker evaluators can access them through the existing workspace mount instead of receiving a temporary path outside the mounted directory.

Workspace state also reports the configured workspace root rather than deriving it from the process working directory.

## Manager MCP readiness

The manager previously could begin processing before its required MCP server and tools had finished initializing. A slow MCP startup could therefore leave the manager without the tools required to control the workflow.

Claude startup now checks the initialization event for the required MCP server and exact tool exposure. Codex startup marks the server as required and detects explicit MCP startup failures.

MCP initialization failures no longer consume the ordinary provider retry budget. NeuriCo retries with progressively larger startup windows until the required tools are available or the operation is explicitly stopped. Provider execution failures continue through the existing provider-failure path.

## HITL progress and execution limits

Full and Auto HITL no longer apply fixed execution timeouts to the resource finder, rule maker, experiment runner, AutoResearch proposer and implementation worker, scorer, or HITL paper writer. Explicit stop handling remains available.

HITL research prompts also no longer introduce an implicit 3,600-second limit when the submitted idea does not define one.

Workers are instructed to keep progressing until the runtime accepts phase completion. If the current approach cannot be executed, the worker must select another valid approach and continue instead of stopping or exiting the phase unfinished.


AndrewRqy:
Thanks a lot for this great work :) The overall recovery design is much stronger, but I found several issues worth addressing before merge:

1. High — dirty nested repositories produce inconsistent scoring inputs. `_materialize_checkpoint_gitlinks` archives the nested repository’s committed Gitlink while preflight and fingerprinting inspect its live working tree. Dirty or untracked files can therefore pass preflight but disappear from the isolated scorer. We should either require nested repositories to be fully clean—including untracked files—and verify HEAD matches the Gitlink, or snapshot the exact live tree used by preflight.

2. High — failed rule-maker repair does not restore the original evaluator. The rollback mechanisms exclude the sealed evaluator paths, so a repair that partially modifies `scoring/eval.py` and then fails leaves the corrupted version behind. I reproduced this locally. The repair flow should keep a runtime-owned immutable backup of the sealed evaluator and restore it exactly on failure or interruption.

3. High — Gitlink archive extraction permits escaping symlinks. `_extract_git_archive` validates archive member names but not symlink targets, then extracts using `filter="fully_trusted"`. A nested repository can contain a link such as `../../.git/config`, which remains an escaping link in the scoring workspace. Symlink and hardlink members should either be rejected or have their resolved targets constrained to the extraction directory.

4. High — permanent MCP configuration failures retry forever. `HitlManager._send` treats every `McpInitializationError` as transient without consuming the normal retry budget. Missing servers, missing tools, or invalid configuration can therefore leave the manager looping indefinitely. Permanent and transient initialization failures should be distinguished, with a bounded retry count or overall readiness deadline for transient failures.

5. Medium — an impossible AutoResearch candidate has no terminal path. The new prompts prohibit rejecting, replacing, or abandoning the candidate, while worker and scorer timeouts are disabled. An infeasible proposal can therefore run indefinitely until an external user stops it. The manager needs a terminal infeasible/irrecoverably-invalid classification plus a configurable attempt or cost ceiling.

6. Medium — the required regression tests are not part of the sealed scoring contract. The manager asks the rule maker to add matching preflight regression tests, but `scoring/tests/` is neither sealed nor copied into the isolated scorer. Those tests can remain agent-visible and then disappear before scoring. We should seal and copy a defined regression-test directory, or explicitly require the checks to live inside `eval.py`.

7. The core recovery paths currently have little regression coverage. The PR adds roughly 1,300 lines across 27 files, but the only changed test file is `test_phase_state.py`. We should add focused tests for repair success/failure/rollback/resume, dirty and malicious Gitlinks, permanent versus transient MCP readiness failures, cooperative stopping with disabled timeouts, and infeasible candidates.

The existing test suite passes locally (`190 passed`), but it does not exercise these new paths, and there are currently no GitHub status checks on the PR.

Frankbest18:
Thanks Andrew — I reviewed each concern and made the following updates:

1. `validate_hitl_rule_maker_outputs()` now validates every Gitlink before the rule-maker stage can be approved. The nested checkout must exist at the recorded repository root, its HEAD must equal the Gitlink commit, and `git status` must be clean, including staged, modified, and untracked files. Isolated scoring can therefore archive the exact commit that passed rule-maker validation.

2. The initial-scoring repair decision is now stored as durable pipeline recovery state before the experiment workspace is rolled back. After the sealed evaluator is restored, the rule-maker stage captures its normal rollback snapshot. Any exception or explicit stop restores that snapshot, including the original evaluator state, while the pending repair record and manager feedback survive for restart. A restarted pipeline re-enters the pending rule-maker repair before running the experiment again.

3. Gitlink archives are now extracted with `tarfile`’s `data` filter instead of `fully_trusted`. This rejects archive links whose targets escape the nested-repository destination. The regression test constructs an escaping symlink and verifies that materialization fails without creating the link.

4. MCP startup-window expiration now has its own `McpReadinessTimeout` classification for both Claude and Codex. Only that condition retries with progressively larger startup windows without consuming the provider budget. Missing tools, invalid configuration, failed handshakes, and other explicit initialization failures remain `McpInitializationError`, consume the existing bounded provider retries, and then stop and roll back an active run through the normal provider-unavailable path.

5. I did not add an agent-declared infeasible terminal state or cost ceiling. This is intentional: an AutoResearch worker must keep working on its assigned candidate rather than deciding independently to abandon it. The explicit terminal mechanisms are user stop and provider unavailability, both of which use the same runtime-owned stop and rollback path.

6. I removed the ambiguous requirement to “add regression tests.” Manager feedback now requires the rule maker to reproduce the concrete failure during the functional `scoring/eval.py` preflight, verify the correction there, and record the evidence in `scoring/rule_maker_log.md`. That log is already part of the sealed evaluator artifacts, so this does not introduce a separate `scoring/tests/` contract.

7. I added focused coverage for clean, dirty, untracked, staged, and mismatched Gitlinks; escaping archive links; repair handoff, interruption, rollback, and restart; and transient versus permanent MCP initialization failures for both providers. The top-level NeuriCo suite currently passes with 222 tests.

AndrewRqy:
Thanks a lot for the detailed fixes! The Gitlink validation, archive-link protection, clarified preflight contract, and additional coverage look good. I found several remaining issues:
1. High — failed rule-maker repair still does not restore the evaluator. HitlStageRollback uses CheckpointManager, which explicitly excludes scoring/eval.py, scoring/targets.json, and scoring/rule_maker_log.md; the private HITL snapshot does not contain them either. I reproduced a failed repair where eval.py remained corrupted after rollback.restore(). The test currently uses a fake rollback and verifies only that restore() was called. We need an immutable runtime-owned evaluator snapshot plus a byte-level restoration test.
2. High — the durable repair request can disappear during recovery. The request is stored in .neurico/pipeline_state.json, but restoring the pre-experiment Git checkpoint can replace that file with its pre-request version. The live process later rewrites its in-memory state, but a crash between those operations loses the repair request. I reproduced a restart immediately after the public restore returning no pending rule-maker repair. The handoff must live outside the restored checkpoint boundary or be preserved transactionally.
3. High — MCP readiness timeouts still retry forever. Permanent initialization failures now correctly consume the provider budget, but McpReadinessTimeout increments only mcp_attempt. If readiness is never emitted, the loop continues indefinitely with progressively larger startup windows. Please add a bounded readiness-attempt count or overall deadline and route exhaustion through the provider-unavailable path.
4. Medium — every nonzero coding-agent exit is classified as provider unavailable. run_prebuilt_cli_agent() currently makes this decision from return_code != 0 alone. Ordinary worker failures, invalid arguments, permission errors, or agent crashes will therefore stop the whole HITL run as provider outages. This terminal path should require a process-launch failure or an explicitly recognized provider/auth/network failure.
5. Medium — extractall(filter="data") is not available across the full declared Python >=3.10 range. Extraction filters were backported in Python 3.10.12 and 3.11.4, so earlier supported patch versions raise TypeError. The Python documentation recommends checking hasattr(tarfile, "data_filter"); the fallback still needs equivalent link-target validation.
The infeasible-candidate behavior is now clearly an intentional policy decision rather than an implementation omission. We should separately confirm that we accept unattended Auto HITL candidates running without an attempt or cost ceiling.
The updated root suite passes locally, but I think items 1–4 remain merge blockers.

Frankbest18:
Thanks Andrew — I went through the second review and updated the branch. I also merged the current `main`, including PR172, and connected its scoring-conformance review to the rule-maker repair path.

1. A rule-maker repair now takes a private snapshot immediately after unsealing the evaluator and before the repair worker starts. That snapshot includes the evaluator, targets, rule-maker artifacts, and private scoring inputs. If the repair fails or is interrupted, those files are restored exactly and any artifacts created by the failed repair are removed.

2. I kept `pipeline_state.json` inside the normal checkpoint. Moving it outside caused pipeline state and workspace state to fall out of sync during rollback, so I reverted that approach. There is now a small control record for the short interval between the manager requesting a repair and the rollback completing. After a restart, NeuriCo uses that record to restore the pending repair into `pipeline_state.json`, then removes it once the repair has been prepared.

3. I kept the unbounded MCP readiness retry as an intentional behavior. It now applies only when Claude or Codex explicitly reports that the current MCP startup window expired. NeuriCo restarts the process with a longer window and waits for the tools to become ready. Missing tools, invalid configuration, failed initialization, and other MCP errors follow the normal bounded provider retry and rollback path.

4. A single nonzero worker exit no longer stops the run. NeuriCo starts a replacement worker and preserves the original process output. A successful replacement clears the failure count. Three consecutive failed provider processes use the existing provider-unavailable stop and rollback path.

5. Gitlink extraction now works safely across the supported Python range. Newer Python versions use the standard `tarfile.data_filter`; older versions use `backports.tarfile`. Both paths reject archive links that escape the destination.

The commits from `2f4ff34` also reduce the amount of frontier history copied into every manager request. Historical attempts now carry their proposal idea ID, manager decision idea ID, proposal type, and acceptance result. The manager can retrieve the full proposal and decision through `hitl-view-ideas`. `view_node` follows the same approach and expands one historical attempt only when the manager requests its proposal ID.

Conversation compaction now considers the size of the complete manager request, including the system prompt, tools, research state, and conversation. Previously it looked only at the conversation, which allowed a large fixed context to exceed the provider limit without triggering compaction.

I also normalized the objective-score record used by initial publication and frontier comparison. NeuriCo now reads both the current score shape and the older nested shape consistently, so restored workspaces and new workspaces make frontier decisions from the same scorer result.

Finally, PR172’s scoring-conformance report now runs in the repair path after rule-maker output validation and before manager review. The report is persisted and replayed during review, and the repaired evaluator is resealed only after that review finishes. The optional rule-maker log is no longer treated as a required repair artifact.

AndrewRqy:
Thanks, the evaluator rollback, repair handoff, and older-Python extraction fixes all look good now.

One issue remains with the new three-strike provider recovery: `run_prebuilt_cli_agent()` returns `provider_process_failed`, but the rule-maker, resource-finder, proposer, and experiment-runner wrappers drop that field when rebuilding their result dictionaries. As a result, the runtime never increments the failure counter, and repeated provider crashes can still launch replacements indefinitely.

Could you propagate `provider_process_failed` through those HITL worker results and add a regression test covering three consecutive failed launches? After that, this should be ready from my side.

Frankbest18:
Thanks, I now carry `provider_process_failed` through the rule maker, resource finder, proposer, and experiment runner when they are running under the Full/Auto AutoResearch runtime. The rule maker and resource finder also preserve the process return code so a successful replacement correctly resets the consecutive-failure count. Ordinary execution paths remain unchanged.

The existing recovery behavior now receives the information it was designed for: the first two provider-process failures launch replacement workers, while the third consecutive failure stops the run through the established rollback path. A result that was already finalized by the runtime is still retained even if the provider process exits nonzero afterward.

AndrewRqy:
thanks for this hard work, im merging this pr

AndrewRqy (approved):

PR #174, opened 2026-09-03
fix(autoresearch): Correct GitHub storage and publication behavior

Frankbest18:
## Summary

This PR fixes GitHub publishing for AutoResearch. The local workspace is now the source of truth throughout the run, and GitHub is treated as an optional storage destination.

Previously, enabling GitHub could change which workspace NeuriCo opened, create ownership against a different path, or pull remote state into the local run. Startup and publication failures could also be hidden, leaving the interface unable to explain why a run did not start or why its results were not published.

## Workspace and repository setup

The AutoResearch launcher now passes its selected workspace directly into the research runner. The runner validates that workspace before acquiring run ownership and uses the same path for fresh and continued runs.

When GitHub publishing is enabled, NeuriCo attaches the workspace to the configured repository without cloning, pulling, or replacing local state. If the repository recorded by the idea has been deleted, NeuriCo creates a new empty repository and publishes the existing local history to it. Repository metadata is written back to the idea’s current registration file.

The detached run worker now loads the project environment before GitHub setup. Selecting GitHub publishing therefore requires valid credentials and organization configuration. Setup failures stop startup and are shown in the interface instead of silently changing the run to local-only mode.

## Publication

A stopped run is no longer published before rollback completes. Successful runs publish the final local checkpoint, including cases where AutoResearch already committed the result and the working tree is clean.

GitHub push results are now checked rather than treating the push call itself as success. Rejected and failed pushes are recorded in the durable launch status, returned as run failures, and presented in the web interface with the underlying sanitized error. The completed local workspace remains available when publication fails.

## Credential boundary

GitHub credentials are now owned by the runtime publishing path. They are removed before starting the manager, coding agents, and provider subprocesses, and are supplied only to the Git process performing the push. The token is no longer written into the workspace’s remote URL.

### DIFF
--- a/src/cli/fetch_from_ideahub.py
+++ b/src/cli/fetch_from_ideahub.py
@@ -657,6 +657,7 @@ def main():
                 idea['idea']['metadata'] = idea['idea'].get('metadata', {})
                 idea['idea']['metadata']['github_repo_name'] = repo_name
                 idea['idea']['metadata']['github_repo_url'] = github_repo_url
+                idea['idea']['metadata']['github_repo_private'] = repo_info['private']
 
                 # Save updated metadata
                 idea_path = manager.ideas_dir / "submitted" / f"{idea_id}.yaml"
--- a/src/cli/hitl_launcher.py
+++ b/src/cli/hitl_launcher.py
@@ -43,9 +43,6 @@ def workspace_for_idea(project_root: Path, idea_id: str) -> Path:
     local_workspace = str(metadata.get("local_workspace", "")).strip()
     if local_workspace:
         candidates.append(Path(local_workspace).expanduser())
-    repository_name = str(metadata.get("github_repo_name", "")).strip()
-    if repository_name:
-        candidates.append(workspace_root / repository_name)
     candidates.append(workspace_root / idea_id)
 
     for candidate in candidates:
--- a/src/cli/hitl_run_worker.py
+++ b/src/cli/hitl_run_worker.py
@@ -276,6 +276,7 @@ def main() -> int:
                 hitl_launch_status_path(Path(str(request["work_dir"]))),
                 {
                     "status": "failed",
+                    "request_id": request.get("request_id", ""),
                     "failed_at": failed_at,
                     "updated_at": failed_at,
                     "mode": request.get("mode", ""),
--- a/src/cli/hitl_run_worker.py
+++ b/src/cli/hitl_run_worker.py
@@ -258,8 +258,6 @@ def main() -> int:
             completed_status,
         )
         control.clear()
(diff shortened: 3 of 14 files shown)


AndrewRqy:
Thanks a lot for this work! The local-workspace-first design and improved publication error handling make sense, but I found several issues worth addressing before merge:

1. High — automatic repository recreation can change its owner and visibility. The recorded `github_repo_url` is ignored, lookup uses only the current GitHub owner and repository name, and any 404 is treated as deletion. The replacement also defaults to public because the original visibility is not retained. An inaccessible private organization repository could therefore be replaced with a public personal repository. I think the safest behavior is to report the access failure and require explicit user action before creating a new publication destination.

2. High — GitHub credential isolation is incomplete for direct CLI runs. The detached worker clears credentials globally, but direct `ResearchRunner` usage does not. The scorer and paper writer still use `os.environ.copy()`, so generated evaluator code and provider subprocesses can inherit `GITHUB_TOKEN`. Credential removal should be enforced consistently at subprocess boundaries.

3. Medium — publication failure produces contradictory status. The worker exits nonzero, but the durable launch record can still say `status: completed` and `success: true`. The nested publication status lets the current UI compensate, but other consumers may treat the run as successful. We should expose unambiguous research and publication outcomes, or mark required publication failure as top-level failure.

The existing suite passes locally (`282 passed`), but this PR currently adds no regression tests for these paths. Focused coverage for repository access/recreation, direct-run credential isolation, and rejected publication would be helpful.

Frankbest18:
Thanks, Andrew. I went through all three concerns and pushed a follow-up.

For repository recreation, NeuriCo now uses the recorded repository URL as the destination identity. It verifies the owner against the configured GitHub owner and retains the repository’s actual visibility alongside its name and URL. A missing repository can only be recreated with that same name and visibility. Incomplete metadata or any owner/name mismatch stops setup with an error. The existing “Publish to GitHub” selection serves as authorization for this exact replacement; a separate visibility control can be considered later as a UI feature.

Credential filtering now happens inside the shared research subprocess launcher. This covers direct `ResearchRunner` use as well as detached runs, including the scorer and paper writer paths that previously copied the parent environment. GitHub credentials remain available only to the runtime publication path.

I also corrected the publication outcome. A completed AutoResearch run now remains successful when GitHub publication fails, since the local workspace is the source of truth. The publication failure and sanitized error remain recorded separately in durable state and visible through the interface, without producing a contradictory nonzero worker exit.

AndrewRqy:
Thanks for the fix, i think this pr is now safe to merge:)

AndrewRqy (approved):

PR #175, opened 2026-09-04
fix(hitl): prevent unattended worker background jobs

Frankbest18:
## Summary

This PR prevents workers from abandoning required background jobs before they finish.

A worker could start a long-running command, end its session, and leave the command behind. NeuriCo would correctly terminate the remaining process, but the replacement worker was not told what happened and could repeat the same behavior indefinitely.

## Changes

- Added a shared rule requiring workers to monitor every job they start until it completes and its outputs are verified.
- Applied the rule to all Full and AutoResearch worker prompts, including replacement workers and proposal generation.
- Added replacement-specific context when NeuriCo terminated jobs left by the previous worker.
- Kept the existing process cleanup and worker-replacement behavior unchanged.

### DIFF
--- a/src/core/hitl.py
+++ b/src/core/hitl.py
@@ -1446,11 +1446,26 @@ class HitlRuntime:
         response = pending.get("response")
         return dict(response) if isinstance(response, dict) else None
 
+    def _replacement_prompt_after_worker_exit(
+        self,
+        prompt_block: str,
+        result: Dict[str, Any],
+    ) -> str:
+        """Report prior job cleanup only when runtime actually performed it."""
+        prompt = str(prompt_block).strip()
+        if not result.get("background_processes_terminated"):
+            return prompt
+        notice = _load_hitl_template(
+            "worker_previous_background_jobs_terminated.txt"
+        ).strip()
+        return f"{notice}\n\n{prompt}"
+
     def _pending_worker_request_replacement(
         self,
         *,
         phase: str,
         worker_name: str,
+        result: Dict[str, Any],
     ) -> Optional[Dict[str, Any]]:
         """Reconnect a replacement worker to one durable held command."""
         from core.hitl_runtime_state import HitlRuntimeState, worker_command_requires_resume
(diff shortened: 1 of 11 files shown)


AndrewRqy:
Thanks a lot for this fix! The shared background-job rule and replacement context are helpful, but I found two remaining issues:

1. High — detected background jobs do not invalidate an already-finalized result. `handle_worker_exit_after_finish()` returns finalized approval before checking `background_processes_terminated`. I reproduced a worker receiving final approval, exiting with a lingering job, having NeuriCo terminate that job, and still returning `approved: true`. Since the terminated job may have left incomplete outputs, background cleanup should override acceptance and trigger rollback or a clean replacement.

2. Medium — the replacement prompt says that all background jobs were terminated, but cleanup only covers the provider’s local process group. Detached processes, Slurm jobs, and remote jobs may continue running without being detected. We should either limit the wording to tracked local child processes or add runtime-owned tracking and cancellation for external jobs.

The focused tests (`12 passed`) and full suite (`287 passed`) pass locally, but they use synthetic result dictionaries. Could you also add a regression test covering a finalized approval with `background_processes_terminated=True`?

Frankbest18:
Thanks, Andrew.

The finalized-approval behavior predates this PR. PR175 does not change the ordering between final approval and background-process cleanup; it only adds the worker instruction and replacement context. I agree that accepting a workspace after terminating unfinished work can be unsafe, but this needs a separate runtime change. The cleanup result does not identify what was terminated, and unconditionally overriding approval could also roll back valid work when the remaining process is the backgrounded `hitl-finish-phase` request itself. I’ll keep that out of this prompt-focused PR and address the finalization boundary separately.

For the replacement prompt, the wording is intentional. Its purpose is to tell the new worker that previous background work is unavailable and must never be relied on. External scheduling and remote-process lifecycle are outside the worker’s responsibility and should not be exposed in its prompt. If an external job can continue modifying the workspace, that is a runtime-isolation issue rather than something this instruction can resolve.

I’m therefore leaving PR175 unchanged.

AndrewRqy:
this looks great:) merging now

AndrewRqy (approved):

PR #178, opened 2026-09-07
Let the HITL manager request resource-finder runs between iterations

AndrewRqy:
The HITL AutoResearch manager can choose which frontier direction to pursue, but resource discovery still runs only at the fixed pipeline entry point. If an iteration exposes a concrete information gap, the manager currently has no way to refresh resources before proposing the next experiment.

This adds a bounded manager-side scheduling API while keeping agent execution on the existing resource-finder HITL path. No second validator, rollback path, or protection system is introduced.

## What this adds

During an eligible frontier-selection boundary, the manager may call `request_agent_run` with `resource_finder`, a focused objective, and a rationale before selecting the next frontier node. Runtime then:

- persists the request separately from frontier-selection state;
- binds it to the selected node inside the existing durable selection transaction;
- runs it before the next proposal through the standard resource-finder plan/execution workflow; and
- checkpoints the resulting resource context so physical attempt retries and later focused refreshes retain it.

The existing HITL worker continuation owns interruption and resume behavior. Existing plan approval, artifact validation, rollback, scoring-file sealing, pipeline-stage tracking, and append-only idea logging remain authoritative. The request tool is withheld when no further proposal will run, and the callable-agent allowlist currently exposes only `resource_finder`.

## Files

| File | Change |
|---|---|
| `src/core/hitl_manager_react.py` / `interactive_manager_tools.yaml` | Expose and constrain `request_agent_run` at eligible selection boundaries. |
| `src/core/hitl_runtime_state.py` | Persist one independent manager-agent request and bind it during the existing frontier decision transaction. |
| `src/core/hitl_autoresearch.py` | Run the bound request before proposal generation and preserve its checkpoint context across retries. |
| `src/core/pipeline_orchestrator.py` / `hitl.py` / `hitl_stage_runtime.py` | Route fixed and dynamic calls through the same resource-finder HITL executor with invocation-scoped approval and logging identity. |
| `src/core/manager_callable_agents.py` | Define the small extensible callable-agent allowlist. |

## Validation

- 11 focused tests cover request availability, durable binding, scoring sealing, retry context, repeated calls, approval identity, and shared execution dispatch.
- Full project test suite: `302 passed`.
- Ruff, compilation, and diff checks pass.


Frankbest18 (changes requested):
Thanks, Andrew. The shared resource-finder workflow and invocation-specific approval identity look aligned with what we discussed. I found the following issues against the intended design:

**[P1] Decouple insertion from frontier selection.**  
The tool’s availability, validation, request identity, and workspace binding all depend on `select_frontier`. Could we move these to an independent runtime-controlled boundary after selection and before proposal, leaving frontier selection unchanged? The current coupling also means repeated, normally completed one-iteration runs never offer the tool.

**[P1] Preserve interrupted resource-finder invocations during recovery.**  
Continuation restores the selected frontier checkpoint before resuming the refresh, losing its updated plan/resources while retaining private continuation state. The inserted stage needs to be recognized by recovery before that restoration occurs.

**[P2] Record the insertion/proceed choice as a manager decision.**  
The request is persisted in mutable runtime state, but the manager’s choice and rationale aren’t recorded as a separate decision in the idea log. Proceeding without resource finding also has no explicit decision record.

**[P2] Retain completed resources independently of the latest request.**  
Refreshing A, then B, then returning to A loses A’s saved resource context because the controller only consults the latest completed request. Could we separate completed-output retention from the pending-request lifecycle?

**[P2] Protect resource checkpoints from Git garbage collection.**  
The context SHA is saved without a retention reference. If no retained descendant protects that checkpoint, it can disappear after reflog expiry and garbage collection while runtime state still references it.

**[P2] Keep sealed verifier artifacts out of public checkpoints.**  
When the evaluator store contains `scoring/verification.json`, unsealing before checkpoint creation commits that protected file into public Git history because it is missing from the checkpoint exclusions. Resealing afterward leaves the committed copy accessible.

The existing 302 project tests passed, but additional regression checks reproduced these cases using real checkpoints/runtime state with stubbed workers. Could we add coverage for them in the revision?

One smaller extensibility concern: `manager_callable_agent_runner(agent, …)` ignores `agent` and hardcodes `resource_finder`. Passing the validated agent through the shared dispatcher would keep the implementation consistent with the tool’s interface.

AndrewRqy:
Thanks a lot for the detailed review!

[P1 insertion boundary] Agent insertion is now its own runtime-controlled decision point after frontier selection and before every proposal. The manager explicitly chooses between proceeding and inserting an available agent. After an insertion completes, runtime returns to the same decision point, so this can support multiple agents later without coupling them to frontier selection.

[P1 recovery] An interrupted inserted-agent run now preserves its in-flight workspace and resumes through the existing HITL worker recovery path instead of restoring the selected frontier over its progress.

[P2 decision log] Both proceed and insert choices, including the manager’s rationale and requested objective, are recorded as B-level decisions in the existing idea log.

[P2 resource retention] Completed resource contexts are retained independently for each frontier node using private Git refs. Returning from A to B and back to A restores A’s latest context, and those checkpoints remain reachable through Git garbage collection.

[P2 sealed verifier] `scoring/verification.json` is now excluded from public checkpoints with the existing protected scoring patterns.

The selected agent is also passed through the shared dispatcher rather than hardcoding resource finder. Added focused coverage for these paths, and the full core suite passes with 306 tests.

Frankbest18 (changes requested):
Andrew, thanks for the update. I found two cases where the new integration breaks behavior supported on `main`.

**[P1] The preparation boundary blocks the first proposal after bootstrap.**

`_prepare_next_proposal()` calls `_latest_frontier_manager_decision_id()` before offering either choice. On `main`, this helper serves frontier maintenance after a scored iteration. The new call also runs before the first proposal after bootstrap, where no previous manager decision is guaranteed. It raises `Frontier maintenance requires a preceding manager decision idea`, preventing continuation.

Could you check this against [`[main](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/hitl_autoresearch.py#L517)`[’s bootstrap behavior](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/hitl_autoresearch.py#L517)](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/hitl_autoresearch.py#L517), [[iteration loop](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/hitl_autoresearch.py#L1300)](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/hitl_autoresearch.py#L1300), and [[existing frontier-maintenance caller](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/hitl_autoresearch.py#L1341)](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/hitl_autoresearch.py#L1341)? The helper’s prerequisite holds at its existing call site, but not at every entry into this new boundary. The same bootstrap continuation reaches its first proposal on `main`.

**[P1] The insertion controller can discard unfinished resource-finder work.**

`_advance_manager_agent_action()` can restore the previous completed resource snapshot when no blocking worker command requires resuming. An unfinished worker can, however, be between commands.

I reproduced this by completing one insertion, starting another, and interrupting that second invocation twice across restarts. On the following restart, the controller restores the first insertion’s snapshot before calling resource finder, removing the unfinished invocation’s changes while retaining its private HITL state. The direct shared resource-finder path preserves the partial workspace in the same test.

Could you review this restore against [`[HitlStageRollback](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/hitl_stage_runtime.py#L125)`[ on main](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/hitl_stage_runtime.py#L125)](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/hitl_stage_runtime.py#L125) and [`[_run_hitl_stage_until_complete()](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/pipeline_orchestrator.py#L313)`](https://github.com/ChicagoHAI/NeuriCo/blob/78940d068c530ec731d4431a8846eca2b122eee2/src/core/pipeline_orchestrator.py#L313)? Those paths coordinate public/private restoration and stage restart. The additional controller restore happens outside that coordination, and the absence of a pending blocking request does not establish that the invocation can safely be reset.

For both cases, I’d like us to preserve the lifecycle guarantees already established on `main` as we add the optional insertion.

AndrewRqy:
Thanks a lot for the detailed review!

[P1 preparation boundary] Bootstrap now records the scored root as evidence when there is no previous manager decision. The premise is persisted with the proposal-preparation action, so recovery reuses the same premise and does not duplicate the decision log.

[P1 insertion recovery] The controller no longer restores the previous resource snapshot based on pending worker commands. In-flight resource-finder work is preserved and recovery is handled through the existing HITL rollback and continuation path, including interruptions between commands.

AndrewRqy:
One follow-up after reviewing the insertion lifecycle against the recovery work in #179:

[P1 invocation identity] Manager-requested agents now use explicit manager_agent provenance with the durable request ID as the invocation identity, instead of overloading AutoResearch attempt provenance.

[P1 interruption recovery] Inserted-agent executions are now treated atomically. An interrupted invocation rolls back through the existing HITL stage boundary and receives at most one automatic retry. The retry count is durable across process restarts.

[P1 bounded failure] Terminal failures are not retried. If an ordinary invocation still fails after its automatic retry, runtime returns to the existing proceed-or-insert boundary so the manager can continue with current evidence or request a fresh invocation.

[P1 context publication] Successful resource contexts now use a durable publication transition. If runtime stops after creating or retaining the context checkpoint, restart completes that transition without rerunning the agent.

Frankbest18 (changes requested):
Thanks for the update, Andrew. I appreciate the work that has gone into addressing the earlier feedback. After going through it again, I still have several concerns that I think need to be resolved before this is ready to merge.

My broader concern is that this feature appears to require only a small routing decision at a controlled AutoResearch boundary: the manager chooses whether to proceed or invoke the existing resource-finder stage, and runtime dispatches that choice. The resource finder’s established HITL lifecycle can then handle planning, review, execution, approval, interruption, and recovery. The current implementation adds another lifecycle around that stage, which creates the following inconsistencies.

**[P1] The optional insertion is still coupled to frontier and proposal-preparation state.**

[`[_prepare_next_proposal()](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_autoresearch.py#L1431-L1641)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_autoresearch.py#L1431-L1641)[[ and ](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_autoresearch.py#L1431-L1641)](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_autoresearch.py#L1431-L1641)[`[_advance_manager_agent_action()](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_autoresearch.py#L1431-L1641)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_autoresearch.py#L1431-L1641) introduce a separate `manager_agent_action` lifecycle with its own statuses, retries, publication phase, checkpoint handling, and recovery behavior. [`[HitlFrontierStore](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_frontier.py#L252-L273)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_frontier.py#L252-L273) is also extended with resource-context refs.

This makes the optional stage part of frontier management rather than an independent stage transition. It also allows the selected frontier SHA to differ from the workspace SHA used for proposal generation, while that relationship is stored only in a private Git ref and is not represented in the public frontier state or idea history.

Could you please compare this with the existing [`[run_plan_centered_hitl_stage()](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/hitl_stage_runtime.py#L64-L157)`](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/hitl_stage_runtime.py#L64-L157), [`[_run_hitl_stage_until_complete()](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/pipeline_orchestrator.py#L580-L609)`](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/pipeline_orchestrator.py#L580-L609), and [`[_run_resource_finder_hitl()](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/pipeline_orchestrator.py#L1293-L1433)`](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/pipeline_orchestrator.py#L1293-L1433)? These already define most of the lifecycle an inserted resource-finder stage needs.

**[P1] An interrupted inserted stage does not follow the established resume path.**

[`[prepare_initial_resume()](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/pipeline_orchestrator.py#L373-L444)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/pipeline_orchestrator.py#L373-L444) treats `manager_agent` provenance separately and restores the rollback boundary. The corresponding [[test expects](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/tests/test_manager_callable_agents.py#L254-L333)](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/tests/test_manager_callable_agents.py#L254-L333) the partial artifact, pending command, and saved continuation to be removed.

Current `main` gives a held request and its saved phase priority during restart. See [`[run_plan_centered_hitl_stage()](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/hitl_stage_runtime.py#L77-L157)`](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/hitl_stage_runtime.py#L77-L157) and [`[_initial_stage_request()](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/pipeline_orchestrator.py#L324-L433)`](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/pipeline_orchestrator.py#L324-L433)[[ / ](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/pipeline_orchestrator.py#L324-L433)](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/pipeline_orchestrator.py#L324-L433)[`[prepare_initial_resume()](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/pipeline_orchestrator.py#L324-L433)`](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/pipeline_orchestrator.py#L324-L433). I think an inserted invocation should retain those same interruption and continuation semantics.

**[P1] A completed resource-finder invocation can run again after a crash.**

In [`[_advance_manager_agent_action()](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_autoresearch.py#L1575-L1637)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_autoresearch.py#L1575-L1637), the action remains `running` while the resource finder executes. After the runner returns success, the durable `publishing` state is written only after another checkpoint is created. If the process stops during that interval, restart still sees `running` and launches the resource finder again.

That can repeat completed searches and tool activity. Reusing the existing stage-completion and restart behavior should help avoid this additional recovery window.

**[P1] Runtime is taking on responsibility for determining evidence and premise relationships.**

[`[_proposal_preparation_premise_id()](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_autoresearch.py#L1675-L1716)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_autoresearch.py#L1675-L1716) searches for a premise and creates a level-C `experiment_runner` evidence record when it cannot find one. [`[log_proposal_preparation_decision()](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl.py#L2231-L2294)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl.py#L2231-L2294) then creates the level-B manager decision and its premise edge. The manager-facing [`[request_agent_run](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_manager_react.py#L1783-L1845)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_manager_react.py#L1783-L1845)[[ and ](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_manager_react.py#L1783-L1845)](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_manager_react.py#L1783-L1845)[`[proceed_to_proposal](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_manager_react.py#L1783-L1845)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_manager_react.py#L1783-L1845)[[ tools](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_manager_react.py#L1783-L1845)](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl_manager_react.py#L1783-L1845) do not let the manager identify the evidence supporting its choice.

My understanding of the established contract is that workers and managers determine the meaning of evidence and decisions, including their premises, while runtime validates and persists what they report. The relevant references are [[HITL ideas and decision levels](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/docs/HITL_AUTORESEARCH.md#L234-L252)](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/docs/HITL_AUTORESEARCH.md#L234-L252), the [[worker idea contract](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/templates/hitl/worker_autonomous_idea_contract.txt#L122-L156)](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/templates/hitl/worker_autonomous_idea_contract.txt#L122-L156), and [`[_manager_phase_premise_id()](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/hitl.py#L1374-L1393)`](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/src/core/hitl.py#L1374-L1393). The manager also already has [`[hitl-view-ideas](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/templates/hitl/interactive_manager_tools.yaml#L72-L82)`](https://github.com/ChicagoHAI/NeuriCo/blob/8650a1d55249daea5b0fffbeaf4e60eb8ccd8294/templates/hitl/interactive_manager_tools.yaml#L72-L82) for examining finalized evidence and premise IDs.

**[P2] The new invocation provenance does not survive idea-record creation.**

[`[manager_agent_provenance()](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/manager_callable_agents.py#L16-L24)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/manager_callable_agents.py#L16-L24) produces `kind` and `invocation_id`, but the idea system currently preserves only [`[parent_node_id](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl.py#L125-L162)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl.py#L125-L162)[[ and ](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl.py#L125-L162)](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl.py#L125-L162)[`[attempt_id](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl.py#L125-L162)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/hitl.py#L125-L162). The new fields are therefore removed when resource-finder ideas are recorded.

As a result, records cannot be reliably associated with the inserted invocation that produced them. Identical reports from separate insertions may be treated as the same idempotent record, and premise lookup may select an idea from an earlier resource-finder invocation.

**[P2] Inserted runs replace the existing pipeline-stage record.**

Every manager-requested invocation calls [`[self.state.start_stage("resource_finder")](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/pipeline_orchestrator.py#L1371-L1384)`](https://github.com/ChicagoHAI/NeuriCo/blob/c78674a125777bf62b82d5c8eb2451b22664d53e/src/core/pipeline_orchestrator.py#L1371-L1384) and later completes the same stage slot. `PipelineState.start_stage()` replaces `stages["resource_finder"]`, so the initial resource-finder run and earlier inserted runs lose their stored outputs, timestamps, and history.

Given how much of the current branch now depends on the additional lifecycle, I wonder whether the cleanest path would be to close this PR, spend some time reviewing the established mechanisms above, and then open a smaller, more focused PR. I suggest this because another patch cycle on the current structure may require more work than returning to the existing abstractions. I believe those references point toward a considerably simpler implementation.

PR #179, opened 2026-09-09
fix(autoresearch): Resume initial runs from their saved stage

Frankbest18:
## Summary

Restarting an interrupted initial AutoResearch run could repeat stages that had already completed. Before the first scored node existed, the launcher selected the fresh path, which checked resource-finder completion but ran the rule maker again.

This PR applies the existing resume mechanisms to the initial run in both Full and Auto modes.

## Changes

Initial runs now check saved stage completion before dispatching a worker. A completed rule-maker stage reuses its sealed evaluator. A completed experiment reuses its recorded score and approval.

For an interrupted stage, NeuriCo reconnects to the matching saved request and continues review or feedback processing. An approval already recorded before the interruption is consumed without launching another worker. When rollback is needed, recovery restores the original stage boundary rather than checkpointing the interrupted state.

Initial-node publication also resumes from its saved progress. An interruption after scoring no longer sends the workspace through experiment execution again.

Explicit evaluator-repair requests retain priority over completed-stage reuse. The change is limited to initial-run resume; existing iteration continuation and Stop behavior remain unchanged.

### DIFF
--- a/src/cli/hitl_launcher.py
+++ b/src/cli/hitl_launcher.py
@@ -140,7 +140,12 @@ class HitlRunController:
                             )
                             stale_request.unlink(missing_ok=True)
             select_hitl_manager_provider(self.work_dir, provider)
-            continuation = HitlFrontierStore(self.work_dir).exists()
+            from core.hitl_autoresearch import initial_publication_requires_resume
+
+            continuation = (
+                HitlFrontierStore(self.work_dir).exists()
+                and not initial_publication_requires_resume(self.work_dir)
+            )
             mode = "continue" if continuation else "fresh"
             request_id = uuid.uuid4().hex
             request = {
--- a/src/core/hitl_autoresearch.py
+++ b/src/core/hitl_autoresearch.py
@@ -264,6 +264,23 @@ def _initial_node_result_from_publication(
     )
 
 
+def initial_publication_requires_resume(work_dir: Path) -> bool:
+    """A partially published initial node is not yet an iteration frontier."""
+    transition = HitlRuntimeState(work_dir).initial_root_publication_transition()
+    return isinstance(transition, dict) and transition.get("status") != "completed"
+
+
+def prepare_initial_hitl_resume(work_dir: Path) -> bool:
+    """Restore initial state before a manager or new worker can act on it."""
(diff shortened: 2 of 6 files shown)


AndrewRqy:
Thanks a lot for working on this fix! The general direction makes sense, but I found two recovery gaps that I think should be addressed before merge:

1. High: resolved raised-idea requests are rolled back instead of resumed. `_initial_stage_request()` only accepts a resolved request when its response has `final=true`. A resolved `raised_idea` normally has no final flag, so it falls through `worker_command_requires_resume()` and is treated as non-replayable even when a valid continuation exists. `prepare_initial_resume()` then restores the stage boundary, losing the manager’s resolution and the worker progress after it. Resolved raised ideas with a matching continuation should reconnect the worker without asking the manager to resolve them again.

2. High: recovered resource-finder approval bypasses artifact validation. `_resume_initial_worker()` converts a saved final approval directly into synthetic success and clears the continuation. The resource-finder path then marks the stage complete without rerunning `resource_artifact_validator()`, so missing or changed `literature_review.md` or `resources.md` can be accepted after restart. The saved approval should be consumed only after the reviewed artifacts are revalidated.

Frankbest18:
Thanks, Andrew. I went through both concerns and compared the initial-run recovery with the existing AutoResearch iteration recovery.

On resolved raised-idea requests, I initially added continuation support, then reverted it. The existing recovery policy rolls back to the saved checkpoint when there is no replayable request. This PR carries that policy into the initial run; I’d prefer to keep a change to that policy separate.

On saved approvals, recovery now checks that the workspace still matches the reviewed fingerprint and reruns the stage’s existing artifact validation before marking it complete. This applies to both resource finding and rule making. I also corrected how those approvals are recognized: the manager’s durable response records approval, but the `final` flag is added later to the worker’s response. Recovery now reads the durable approval directly and distinguishes stage completion from plan approval.

While tracing these paths, I found another restart issue shared by resource finding and rule making. An approved plan could send the replacement worker into execution even though its unresolved request belonged to review. The shared stage executor now restores the saved phase and reconnects to the original request without overwriting its continuation.

AndrewRqy (approved):

PR #180, opened 2026-09-10
Let HITL managers insert resource-finder runs before proposals

AndrewRqy:
The HITL AutoResearch manager can choose which frontier direction to pursue, but resource discovery still runs only at the fixed pipeline entry point. If an iteration exposes a concrete information gap, the manager currently has no way to refresh resources before proposing the next experiment.

This adds a separate proposal-preparation decision where the manager may either proceed or request a resource-finder run. The addition reuses the existing HITL stage lifecycle rather than introducing a parallel invocation or recovery system.

## Architecture

```text
Frontier selection
        ↓
Proposal-preparation decision
        ├─ proceed ───────────────→ Existing proposal pipeline
        │
        └─ insert agent
                ↓
        Existing HITL agent executor
                ↓
        Proposal-preparation decision again
```

The manager boundary owns only the routing choice. The selected agent is validated and dispatched into its existing HITL executor, which remains responsible for planning, review, execution, approval, interruption, rollback, and recovery. After an inserted run completes, runtime returns to the same decision boundary so the manager may insert another available agent or proceed.

The feature intentionally does not introduce a separate manager-agent lifecycle, retry system, workspace publication path, validation layer, or rollback mechanism. Adding another callable agent should require registering it and routing it to its existing standardized executor—not creating another execution model.

## What this adds

Before each physical proposal attempt, the manager may call `request_agent_run` with `resource_finder`, a focused objective, a rationale, and the premise idea supporting the request. Runtime then:

- persists the choice through the existing next-action state;
- executes the request through the standard plan-centered resource-finder HITL path;
- preserves the normal approval, continuation, rollback, and restart behavior;
- seals scoring resources while the inserted agent runs; and
- returns to the proposal-preparation boundary after the run completes.

The decision point is independent from frontier selection so it can later support additional callable agents without coupling their lifecycle to proposal selection. Existing stage records now retain invocation history so repeated resource-finder runs do not overwrite one another.

## Files

| File | Change |
|---|---|
| `src/core/hitl_manager_react.py` / `interactive_manager_tools.yaml` | Add the proceed-or-insert manager decision and validate its evidence. |
| `src/core/hitl_autoresearch.py` | Place the decision before each physical proposal attempt and preserve it across resume. |
| `src/core/pipeline_orchestrator.py` / `hitl.py` / `hitl_stage_runtime.py` | Route fixed and inserted runs through the same HITL stage executor and existing provenance semantics. |
| `src/core/hitl_runtime_state.py` | Reuse durable next-action state and retain repeated stage history. |
| `src/core/manager_callable_agents.py` | Define the small extensible callable-agent allowlist. |

## Validation

- 15 focused tests cover decision persistence, premise validation, retry boundaries, resume behavior, stage history, and shared execution dispatch.
- Full core test suite: `306 passed`.
- Compilation and diff checks pass.


Frankbest18 (changes requested):
Thanks for reworking this. The overall control flow now looks aligned with the intended design. Returning to the same decision boundary also supports multiple resource-finder runs cleanly.

I found one blocking issue and several smaller consistency problems:

**[P1] Preserve completed resource-finder output across proposal rollback**

`_prepare_next_proposal()` runs the inserted stage before `run_iteration()`, but the subsequent proposal attempt continues to use the original selected frontier `parent_sha` as its rollback target. If the proposal fails, candidate publication is incomplete, or the candidate is rejected, runtime restores that original checkpoint. This removes the updated `resources.md` and `literature_review.md`.

This can happen during an ordinary rejection, without a crash:

1. The manager requests resource finding.
2. Resource finding completes and updates the public artifacts.
3. The manager proceeds to proposal.
4. The candidate is rejected.
5. `_restore_rejected_candidate_workspace()` restores the original frontier checkpoint, discarding the resource-finder work.

The private resource-finder ideas and preparation decisions can remain after this restore, leaving the research record associated with public artifacts that have reverted to older contents. Multiple consecutive resource-finder runs can all be lost in the same way.

There is also a restart window after a completed preparation action is cleared and before the next preparation action or physical proposal attempt is opened. At that point, startup sees no pending preparation action and restores the selected frontier.

A completed inserted stage should establish a durable proposal-preparation baseline. Later candidate rollback should return to that baseline rather than removing work completed before the proposal began.

References:

- `src/core/hitl_autoresearch.py::_run_iteration_until_scored`
- `src/core/hitl_autoresearch.py::_prepare_next_proposal`
- `src/core/hitl_autoresearch.py::run_iteration`
- `src/core/hitl_autoresearch.py::_restore_rejected_candidate_workspace`
- `src/core/hitl_manager_react.py::begin_proposal_preparation`

**[P2] Keep insertion identity separate from&#x20;****`attempt_id`**

`log_proposal_preparation_decision()` stores the preparation `invocation_id` as `attempt_id`, and the same substitution is used for resource-finder provenance.

On `main`, `attempt_id` identifies a physical AutoResearch proposal/experiment attempt and corresponds to its attempt history and rollback transaction. Proposal preparation occurs before that attempt exists. The insertion needs its own identity for replay, approval scoping, and repeated runs, but that identity should remain an `invocation_id` rather than changing the meaning of the established `attempt_id` field.

References:

- `src/core/hitl.py::log_proposal_preparation_decision`
- `src/core/pipeline_orchestrator.py::_run_resource_finder_hitl`
- `src/core/autoresearch_common.py::attempt_id_for`
- `src/core/hitl_autoresearch.py::run_iteration`

**[P3] Record proposal preparation as its own phase rather than a frontier phase**

`begin_next_autoresearch_action()` maps the new action to `preparing_proposal`, but `_record_phase_transition_unlocked()` still receives `stage="frontier"`. The workspace UI therefore presents this decision as “Frontier” activity even though frontier selection has already completed.

The execution flow is properly separated now, so the persisted phase and UI label should reflect that same separation.

References:

- `src/core/hitl_runtime_state.py::begin_next_autoresearch_action`
- `src/core/hitl_runtime_state.py::record_next_autoresearch_action_decision`
- `src/core/hitl_workspace_view.py::_phase_notification`

**[P3] Report preparation-boundary manager failures accurately**

`_cancel_backend_failed_runtime_request()` describes every runtime action other than `select_frontier` as “frontier pruning.” A manager backend failure during `prepare_proposal` therefore reports that the frontier-pruning boundary failed. Please give `prepare_proposal` its own boundary description.

Reference:

- `src/core/hitl_manager_react.py::_cancel_backend_failed_runtime_request`

The existing tests cover tool validation, decision replay, repeated stage records, and resuming a completed invocation, but they do not exercise a successful inserted run followed by candidate rejection or the post-completion restart window. A regression test covering those workspace transitions would help verify the required ownership boundary.

With the P1 issue addressed, the main architecture should be in good shape; the remaining items are localized consistency fixes.

AndrewRqy:
Thanks, addressed all four issues in the latest commit.

Completed manager-invoked resource-finder runs now create a standard checkpoint and update the selected frontier node’s workspace baseline, so later rejection, rollback, selection, and restart paths preserve the completed resource work.

I also added a dedicated invocation_id field throughout the existing runtime and idea provenance flow. This identity is needed to distinguish replay of an interrupted invocation from a later identical request, while keeping it separate from attempt_id, which remains reserved for physical AutoResearch proposal/experiment attempts.
Proposal preparation is now reported under the experiment-runner stage, and backend failures identify the preparation boundary correctly.

Added regression coverage for prepared-workspace restoration, pending and completed invocation recovery, repeated invocation identity, and separation from attempt semantics.

Frankbest18 (changes requested):
Thanks for the update, Andrew. I reviewed the current branch against `main` again. The phase-label fixes look appropriate, but there are still several issues that need to be addressed before merging.

I also need to correct my earlier feedback about insertion identity. I correctly flagged the use of `attempt_id`, but I should have questioned whether a separate invocation identity was needed before recommending it. The original version already generated that identity; the latest update expands it into global provenance. That expansion is not justified by the existing action, worker-request, continuation, and rollback mechanisms.

**[P1] Random invocation identity changes logical idea identity**

Adding `invocation_id` to the idea-record payload makes two otherwise identical manager decisions produce different idea nodes solely because their random tokens differ. The new repeated-decision test explicitly requires this behavior.

`HitlIdeaLog.append(idempotent=True)` on `main` compares logical content, including premises. A later decision based on resource-finder results can cite that new evidence and therefore have a different premise. Runtime occurrence identity should not create that distinction on the manager’s behalf.

Please preserve the existing separation between runtime replay state and logical idea content. The relevant references are `HitlIdeaLog.append()`, `HitlIdeaLog.logical_payload()`, and the `HitlRuntimeState` next-action lifecycle.

**[P1] Accepted frontier nodes can restore a different, unscored workspace**

`update_workspace_checkpoint()` attaches a new checkpoint `W` to an accepted frontier node `P`. The node’s score, plan, acceptance reason, and history still describe `P`, but selection, continuation, rejection, and recovery can restore `W`.

Resource finder can modify datasets, code, dependencies, and other public files. Those changes have not been evaluated as part of `P`, so treating `W` as the workspace of the scored node detaches the score from its evaluated contents. The next candidate can also have Git parent `W` while its recorded frontier parent remains `P`.

The new test explicitly accepts `current_best_sha == P` while `HEAD == W`. That conflicts with the established frontier contract.

The same method mutates the accepted node record and retargets `refs/neurico/hitl/frontiers/{P}` to another commit. Please preserve the relationship between an accepted node, its immutable record, its ref, and its exact scored checkpoint.

References: `HitlFrontierStore`, `update_workspace_checkpoint()`, `HitlAutoResearchController.run_iteration()`, and the frontier invariants in `docs/HITL_AUTORESEARCH.md`.

**[P1] Recovery reports a checkpoint it did not restore**

The recovery paths now restore workspace checkpoint `W` but can return logical node SHA `P` as `restored_checkpoint_sha`. The CLI then publishes that value as the restored checkpoint. Continuation similarly reports `current_best_sha=P` while the workspace is at `W`.

These results need to describe the actual restored state. This is another consequence of giving an accepted frontier node a second workspace identity.

References: `recover_interrupted_hitl_attempt_if_needed()`, `continue_hitl_autoresearch()`, and `src/cli/hitl_run_worker.py`.

**[P1] Private frontier state and public history can disagree**

`update_workspace_checkpoint()` modifies the private node without updating its public history mirror. If execution stops after resource finding and before another candidate is finalized, the private record says that `P` restores `W`, while the public record still describes the original `P`.

Please preserve the existing publication guarantees rather than introducing a separate private update path. See `HitlFrontierStore.mirror_nodes_to()` and the root and candidate publication paths in `hitl_autoresearch.py`.

**[P1] The inserted stage unseals the evaluator between attempts**

The inserted-agent runner calls `unseal_scoring_files()` in its `finally` block. HITL AutoResearch normally retains the verified sealed evaluator between attempts.

This returns the evaluator and private test resources to the public workspace and removes the authoritative sealed copy while the manager decides what to do next. If those public files change, a subsequent immutable seal can accept the changed versions as authoritative.

The resource-finder call should preserve the existing scoring-seal lifecycle. See `scoring_seal.py` and the HITL scoring setup in `hitl_autoresearch.py` and `pipeline_orchestrator.py`.

**[P1] The default API path creates separate managers for the same session**

When no manager is injected into `run_hitl_autoresearch_loop()`, the experiment runtime creates a default manager and the nested `ResearchPipelineOrchestrator` creates another.

They maintain separate cached conversations over the same durable transcript. After the nested manager writes the resource-finder review context, the outer manager can overwrite it with its stale conversation at the next preparation turn.

Please pass the session’s existing manager into the inserted stage. The relevant paths are `run_hitl_autoresearch_loop()`, `ResearchPipelineOrchestrator`, and `HitlManagerTranscript`.

**[P2] Invocation identity spreads into approval and recovery semantics**

The random identifier also affects approval scopes, worker requests, continuation records, rollback provenance, stage records, and log names. Recovery falls back from `invocation_id` to `attempt_id`, despite their supposedly separate meanings.

On `main`, plan approval is tied to the pipeline stage and exact plan fingerprint. Adding a random invocation to that scope requires another approval for identical content simply because another call occurred. A changed plan already produces a different fingerprint.

Please review this against `mark_plan_approved()`, `has_plan_approval()`, `run_plan_centered_hitl_stage()`, and `HitlStageRollback` before introducing another identity across these systems.

**[P2] Optional runs overwrite the fixed Stage 1 record**

Repeated inserted runs reuse `PipelineState.stages["resource_finder"]`, which existing startup and resume code interprets as the fixed first pipeline stage.

The new `stage_history` list preserves older records, but the existing consumers still inspect the active slot. It therefore does not resolve the conflicting meanings of that field. It also changes restart behavior for every pipeline stage without updating the corresponding status and recovery consumers.

Please keep optional invocation state separate from the fixed Stage 1 record. See `PipelineState`, `resume_pipeline()`, and the fixed-stage checks in `runner.py`.

**[P2] Fixed Stage 1 dispatch is unnecessarily coupled to optional dispatch**

The established resource-finder Stage 1 path now goes through `run_hitl_agent_stage_until_complete()` and the optional-agent registry.

Exposing an optional manager call does not require changing how the fixed stage is dispatched. Please retain that existing path and keep the new routing boundary independent.

**[P2] Initial-stage recovery now represents post-initial work**

The changes to `_initial_stage_request`, `_stage_rollback`, `_resume_initial_worker`, `_completed_initial_stage`, and `prepare_initial_resume` extend initialization recovery to inserted work between AutoResearch attempts.

Using the `"initial_stage"` recovery slot for that work changes its established meaning. Please trace the existing resource-finder continuation and rollback lifecycle and reuse it without redefining the initialization boundary.

**[P2] Frontier restoration is skipped without checking the preparation state**

The presence of any `prepare_proposal` action suppresses selected-frontier restoration. This includes pending, proceed, cancelled, and resolved actions, rather than specifically identifying an active inserted operation whose workspace must be preserved.

Any restoration exception needs to follow a precise durable execution state. Please check the conditions in `continue_hitl_autoresearch()` and `HitlAutoResearchController.run()` against the existing next-action transitions.

**[P2] The tests and documentation do not establish compatibility with `main`**

The 312 passing core tests do not resolve these concerns. Some new tests explicitly require random tokens to distinguish identical ideas, a reported frontier SHA different from `HEAD`, and optional calls overwriting the fixed-stage slot.

The tests should instead verify the existing contracts: sealed scoring resources, exact frontier identity, consistent public and private history, accurate recovery results, preserved manager context, and unchanged fixed-stage resume behavior.

`docs/HITL_AUTORESEARCH.md` also still describes immutable accepted checkpoints. The mutable workspace pointers and altered recovery semantics conflict with that model.

The manager’s proceed-or-request boundary is a useful addition. The concerns above are with the changes to established mechanisms around it, and those need to be resolved before this feature can be merged.

Frankbest18:
Thanks for your work on this, Andrew. Given the scope of the issues above, I suggest closing this version and opening a new draft PR from current `main`. A fresh start would make it easier to review the feature without carrying forward the additional mechanisms introduced during these revisions.

I would appreciate your continuing to lead the implementation. It may help to first trace the existing manager-decision and resource-finder lifecycles, particularly idea logging, approvals, interruption, recovery, and preservation of completed work. The references in the review above should be useful starting points.

A brief outline of which existing components will support the feature would also help us agree on the approach early, including how repeated resource-finder calls and interrupted runs would be handled.

One thing I have found useful in my own development is running a few local smoke tests with real research ideas before submitting a PR. Exercising the actual manager and worker harnesses—through resource finding, proposal, rejection, and an interrupted run—often reveals visible problems that isolated tests miss. Looking through the resulting artifacts and idea log can be particularly helpful. A successful run does not establish that the design is sound, but it gives another useful check on whether the workflow behaves as intended.

The manager decision boundary and the two tool choices are useful work to carry forward. I think building from the existing mechanisms on `main` will give us a smaller implementation that is easier to review and maintain.

PR #181, opened 2026-09-11
feat(ordinary): Add managed ordinary research in Full and Auto modes

Frankbest18:
## Summary

This PR brings NeuriCo’s ordinary research workflow into the managed research interface.

Ordinary research keeps its existing two-stage structure: resource finding followed by experiment execution. It can now run in either Full mode, where the manager escalates human-level decisions, or Auto mode, where the manager resolves them directly. It does not introduce rule making, objective scoring, frontier management, or AutoResearch iterations.

## Managed ordinary research

A new managed ordinary entry path connects the existing pipeline to the same workspace ownership, detached execution, manager review, Stop, rollback, provider-failure handling, publication, and paper-writing mechanisms used by AutoResearch.

The resource finder and experiment runner use the established plan, execution, and review workflow. Interrupted runs reuse durable stage state and pending requests. Stopping restores the active stage boundary, and completed stages are reused when the workspace is started again.

Ordinary runs are kept separate from AutoResearch. The selected workflow is recorded in the workspace when execution begins, and later launches cannot switch that workspace to the other workflow. Existing AutoResearch workspaces and older launch requests retain their current behavior.

## Web and CLI

The run configuration now lets users choose between AutoResearch and Ordinary research independently of the Full/Auto setting. Iteration count is shown only for AutoResearch.

Both interfaces display the active workflow and mode from durable workspace state. Ordinary workspaces omit the Nodes and Whiteboard views because those concepts belong to AutoResearch. An unsuccessful ordinary run is shown as incomplete instead of complete, so it can be resumed from its saved stage.

Manager instructions related to evaluators and objective scoring are now included only when scoring is active. Ordinary experiment reviews therefore remain focused on the submitted research plan and completed work.


AndrewRqy:
One issue:

The managed ordinary `experiment_runner` no longer preserves the existing output contract. The standard ordinary workflow requires `REPORT.md`, but this path starts the stage without expected outputs and uses no non-scoring artifact validator. As a result, the manager can approve the stage and the pipeline can report success even when no report was produced.

Could we reuse the existing ordinary `REPORT.md` validation here so managed execution retains the same completion requirements as the original workflow?

Frankbest18:
Thanks for catching this. The managed ordinary path was bypassing the standard pipeline-state validation, while its replacement artifact validator was only enabled for scored AutoResearch.

Managed ordinary execution now retains the existing `REPORT.md` contract at both boundaries: runtime checks for the report before manager review, and pipeline state records it as a required output for completion and recovery. The scored AutoResearch path is unchanged.

AndrewRqy (approved):

PR #182, opened 2026-09-15
feat(autoresearch): Construct managed baselines from Ordinary research

Frankbest18:
## Summary

This PR adds a manager-mediated workflow for converting completed Ordinary research into an AutoResearch baseline. It is stacked on `ordinary-neurico-upgrade`.

The existing bootstrap commands remain unchanged. The new workflow is available through the HITL web and CLI interfaces and supports both Full and Auto decision modes.

## Baseline construction

Baseline construction is available only after an Ordinary run has completed successfully and before an AutoResearch frontier exists.

The runtime creates a deterministic, value-redacted manifest of the completed experiment outputs. The rule maker uses the original idea, resource material, and this structural manifest to select the authoritative research output and construct the scoring contract. It cannot inspect the completed output values or modify the completed experiment.

The rule maker then proceeds through the existing managed plan, execution, validation, and manager-review flow. Changes outside the scoring contract are rejected and restored through the existing stage recovery mechanism.

Once the evaluator is approved, the runtime seals it and scores the unchanged completed workspace in isolation. The manager reviews the runtime-derived result and either approves the baseline or returns a concrete evaluator correction to the rule maker. This workflow never routes scoring feedback to the experiment runner because the completed experiment is fixed.

Approval publishes the scored workspace as the initial AutoResearch frontier node and preserves the constructed evaluator as the workspace’s scoring contract. The workspace can then continue through the existing AutoResearch workflow.

## Output selection

The earlier implementation delegated output selection to the manifest trimmer. That made baseline construction depend on a model-generated manifest and allowed malformed output to fall back to an overly broad mechanical selection.

Output discovery is now deterministic. The runtime scans the completed workspace, removes runtime-owned and unsuitable artifacts, and exposes only value-redacted structural information to the rule maker. The rule maker remains responsible for choosing which eligible output represents the research result, while the manager reviews that choice before scoring.

The selected output contract is persisted with the completed baseline so later AutoResearch runs use the same approved artifact paths.

## Interfaces

The web and CLI launch flows expose `Construct AutoResearch baseline` only for eligible Ordinary workspaces. Run status identifies baseline construction separately from Ordinary research and AutoResearch.

Provider selection, Full or Auto mode, GitHub publication, run ownership, detached execution, stopping, recovery, and renderer behavior continue through the existing managed-run infrastructure.


AndrewRqy:
Thanks a lot for working on this! The managed baseline construction direction makes sense, but I found two recovery gaps that I think should be addressed before merge:

1. High: stopping during baseline construction uses the wrong recovery configuration. `_finalize_stopped_run()` identifies the workspace as ordinary research, but creates the recovery orchestrator without `baseline_construction=True`. The active `rule_maker` boundary is therefore rejected as an unknown initial stage, and Stop reports a recovery failure instead of restoring the workspace. Stop recovery should configure the orchestrator from the requested operation.

2. High: an interruption during the baseline rule-maker leaves the public and private recovery state inconsistent. The pre-publication rollback restores the public checkpoint and removes the `initial_stage` rollback descriptor, while the pending worker request remains in `.neurico/hitl`. The next launch finds the pending request without its original rollback boundary and cannot resume or restart it. The active managed stage should be reconciled through the existing `HitlStageRollback` mechanism before the outer pre-publication checkpoint is restored.**

